PR #10, opened 2018-10-11
require: bump django-filter to 2.0.0

dimidd:
Solves the import error
https://github.com/carltongibson/django-filter/issues/956

Signed-off-by: Dimid Duchovny <dimidd@gmail.com>


icoxfog417:
Thank you for your contribution!

PR #18, opened 2018-10-17
Feature/auto labeling

jamesmf:
This adds a few pieces of functionality:
- Lets you autolabel an entire project with a given model using a new command `python manage.py autolabel <project_id> <model_str>`
- Adds an endpoint for "confirming" that you agree with the automatically generated labels
- Hits that endpoint when Vue calls nextPage()
- Simple support for a "seed model" that does simple flexible matching on some `seed_data.json` that looks like `{'entity_type1': [seed_term1, seed_term2], 'entity_type2': [seed_term3, seed_term4],....}`

I'd love to hear thoughts on whether this is worth supporting (avoids the need to host a model as a separate service) and if so how it can be refactored.

Things that would make this more useful:
- Downloading only docs where `manual=True`
- Resetting all `manual=False` annotations when re-calling `python manage.py autolabel`
- Working with the feature request to export sequence annotations as entity spans (instead of per-character)


jamesmf:
I think functionally it's nice to add this as a Command, but I think it's important to store some kind of Review or Confirmation object separate from the Annotation. I am thinking we could add a model for one of those and have the ConfirmAnnotation function hit it on nextPage().

The logic is that there are instances where you want to confirm that there are no annotations in your document, so the logic can't live in an Annotation.

I would also love to refactor the `load_model` logic so that it isn't spacy-specific

jamesmf:
Let me know if there's any way I can refactor this to make it a candidate for merging back into this branch.

PR #25, opened 2018-10-26
Feature/delete document

jamesmf:
Added an API endpoint to delete a document and a button just above the `Prev` button. Works fine in my limited testing.

The offset-matching-regex could probably be broken if someone were to search with a specific query string, but seems unlikely.


tugstugi:
@Hironsan is it possible to delete a document?

Issue #28, opened 2018-11-08
can't annotate two neighbor entity in a sentence?


phiedulxp:
```
can:
xx ORG xx ORG xx
can't:
xx ORG ORG xx
```

BrambleXu:
I test your case with my text locally, and it works well. Could you give us more detail about your data?

phiedulxp:
> I test your case with my text locally, and it works well. Could you give us more detail about your data?

it only happen in language which has no blank space between words.
这个工具不是未中文设计的！
如果分词后再标注，错分时需要跨空格划词。

BrambleXu:
Thanks for the demo.  I reimplemented it.  I will work on this problem.

Hironsan:
Solved by https://github.com/chakki-works/doccano/pull/30.

Thanks @BrambleXu !

PR #29, opened 2018-11-12
solve issue 28 for neighbor_labeling

BrambleXu:


BrambleXu:
#28

Hironsan:
Thank you for your great contribution.
But there is still a bug.

For example:

* there is an annotated data ranged from 4 to 7.
* I want to annotate text span from 1 to 9.

In such case, I can annotate the text but actualy the case is invalid.

PR #30, opened 2018-11-12
solve issue 28 for neighbor_labeling and contain label annotation problem

BrambleXu:
Two bugs are solved.
The first bug is we can not annotate two words if they are neighbors, especially in the non-space language like Chinese and Japanese.  #28 
The second bug is if we annotate a new label, this new label can contain some label already exist.


Hironsan:
Thanks! But there is still a bug.

For example:
* There is an annotation ranged from 4 to 7.
* When I annotate a text span from 4 to 7.

This is an invalid case but I can annotate it. :)

Issue #31, opened 2018-11-15
Always delete the latest project


studentJiao:
I'm attempting to delete a project, but it's always that the latest project was deleted.

BrambleXu:
Thanks for your feedback, I confirmed this situation.

PR #32, opened 2018-11-15
solve issue #31 for project deletion problem

Hironsan:
A bug a fixed in #31.
We can delete any project.


BrambleXu:
Thanks for your rapid response :)

Issue #17, opened 2018-10-16
Support importing datasets in JSON format


Zerthick:
Currently in order to upload a dataset for annotation all of the documents in the dataset must be in a single txt file with each document separated by a newline character like so:
```
document 1 text
document 2 text
```
It would be very helpful if instead documents could be within separate files and a bulk upload could be performed.  For example:

**Document 1**
```
Document
1
text
```

**Document 2**
```
Document
2
text
```

This would also allow documents to contain newline characters which was requested here: #12 

Love the tool!

Zerthick:
Alternatively, supporting uploading a single file in JSON format such as:
```JSON
[
"Document 1 text",
"Document 2 text"
]
```

Would also work.

Zerthick:
I've created a bare-bones implementation of this second approach utilizing a single JSON file here:
https://github.com/Zerthick/doccano/pull/1

It currently adds a second radio button to allow the user to specify which format file they would like to upload, however this could be changed to infer the format based on the file name extension.

In addition, this method of upload allows documents to contain both newline characters and commas within their text, meaning the rendering of the annotation views should support this behavior.

icoxfog417:
Thank you for your proposal! Certainly supporting one-data-one-file format (like jsonl) is good for the data including newline or complex structure.

When supporting this feature, I think it will be good to enable uploading archived file (like `tar.gz` or `zip`). We contine to improve the way to upload the data.

Zerthick:
I agree, uploading a `tar.gz` archive would be helpful as well.

Zerthick:
So is there a consensus on the format of the JSON file? I can update the parser as needed.  Also, is the separate radio button an acceptable way to implement this functionality in the UI, or would we want to auto-detect based on file extension?  Finally, if the user uploads a zip or tar file should it be expected that the archive contains a single file and if so, what is the format of that file, CSV, JSON, both?

BrambleXu:
@Zerthick, thanks for your advice. We had a discussion about your request and I will share it with you. 

* So is there a consensus on the format of the JSON file? 
  * Yes, we want to support the JSON file. 

* Also, is the separate radio button an acceptable way to implement this functionality in the UI, or would we want to auto-detect based on file extension? 
  * The radio button is a great idea, we will accept it.

* Finally, if the user uploads a zip or tar file should it be expected that the archive contains a single file and if so, what is the format of that file, CSV, JSON, both?
[...]

PR #33, opened 2018-11-27
iss17: support for uploading json file

BrambleXu:
#17 
* Add json upload button and format value in `dataset_upload.html` 
* Change `views.py` for processing json file
* Add test case for json upload feature


BrambleXu:
Thanks for the review, working on it!

Issue #27, opened 2018-11-02
Adding in linebreaks filters


Zerthick:
I'm attempting to add in linebreak filters to allow newlines to be rendered within documents to complement #17 however I can't seem to get them working.

I've identified the line which inserts the document text to be line 34 of document_classification.html
https://github.com/chakki-works/doccano/blob/master/app/server/templates/annotation/document_classification.html#L34

However modifying it to `[[ docs[pageNumber].text|linebreaks ]]` does not produce the desired effect.

On a side note I am not very familiar with Django, what is the significance of the double square-bracket syntax? I have not seen it in any Django examples.  Aren't variables typically inserted with double curly-brackets in templates?

Zerthick:
I've managed to address the issue by modifying the line to:
```
<span style="white-space: pre-wrap;" id="myspan">[[ docs[pageNumber].text ]]</span>
```

I've noticed you and an issue with vue regarding the spacing of in-line content. Vue is preserving the spacing of the content within the template so changing the above line to something like

```
<span style="white-space: pre-wrap;" id="myspan">
    [[ docs[pageNumber].text ]]
</span>
```

Causes the text to become indented when rendered in the html page.

BrambleXu:
Hi @Zerthick, you mentioned document_classification.html, which means you create a document classification project? Could you give me more detail about what project you use, what file format you upload, and the problem you meet. It would be helpful if you could share your demo json file.

BrambleXu:
Thanks for your code, @Zerthick. I test it and it works well for document classification task and seq2seq task while annotating. But the exporting csv file will render the newline and this will cause one sentence seperate to two lines. In order to avoid this situation, I will provide the option to export json file. #15

Zerthick:
@BrambleXu Added some more comments to https://github.com/chakki-works/doccano/pull/34

BrambleXu:
Thanks, @Zerthick. The information you provide is very useful!

BrambleXu:
Solved in #37

BrambleXu:
Hi, @Zerthick. In the upcoming JSON export feature, even if we won't rendering line breaks in annotation page for sequence labeling task, the exported JSON file will contain linebreaks.

PR #34, opened 2018-11-28
Enhancement/linebreak

BrambleXu:
* iss27: render linebreaks in document classification and seq2seq tasks
* enhancement: add button to jump between edit data page and annotation page

### DIFF
--- a/app/server/templates/admin/admin_base.html
+++ b/app/server/templates/admin/admin_base.html
@@ -4,6 +4,12 @@
 <link rel="stylesheet" href="{% static 'css/admin.css' %}">
 {% endblock %}
 {% block navigation %}
+<a class="navbar-item" href="{% url 'annotation' view.kwargs.project_id %}">
+  <span class="icon">
+    <i class="fas fa-user-edit"></i>
+  </span>
+  <b>Annotate Data</b>
+</a>
 <a class="navbar-item" href="{% url 'upload' view.kwargs.project_id %}">
   <span class="icon">
     <i class="fas fa-upload"></i>
--- a/app/server/templates/annotation/annotation_base.html
+++ b/app/server/templates/annotation/annotation_base.html
@@ -5,6 +5,14 @@
 <script src="https://unpkg.com/marked@0.3.6"></script>
 <script src="https://unpkg.com/lodash@4.16.0"></script>
 {% endblock %}
+{% block navigation %}
+<a class="navbar-item" href="{% url 'dataset' view.kwargs.project_id %}">
+  <span class="icon">
+    <i class="fas fa-edit"></i>
+  </span>
+  <b>Edit Data</b>
+</a>
+{% endblock %}
 {% block content %}
 <div class="columns" id="mail-app" v-cloak>
   <aside class="column is-3 aside hero is-fullheight">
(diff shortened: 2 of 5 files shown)


BrambleXu:
Right now we have no intent to support rendering linebreaks in sequence labeling tasks.

BrambleXu:
The reason we removing linebreaks is not only the readering problem, but also the csv format output will be out of aligment. 
8c90c071-5776-47b6-804c-4dbe0558c885

As for redering linebreaks in sequence labeling task, I will find time to talk with @Hironsan in detail. 

Your caution is reasonable. Could you share with us what kind of task and data you use, and how the linebreaks can provide important information? 
We can take the conversation in the issue #27 page.

BrambleXu:
I tested the NER training case with Spacy. If the text contains line breaks, the annotation would not be recognized.

This is the case with line breaks.
e85fb875-ab20-48cc-89ab-379bf10d1200

ff09f8fd-9e77-430c-9a82-c457ad1bb41c

This is the case without line breaks.
c3a15edd-d42d-435f-8682-536dff66d696

951e280b-390b-4269-b762-3657c33180f8

So for the sequence labeling task, the output text can not contain line breaks. And due to this reason, there is no need to render the line breaks while annotating.  I think deleting line breaks is a suitable way to get the correct export format. 

@Zerthick Do you agree with that? Welcome any opinions.

Zerthick:
The line breaks count as an additional character so you would need to update the indexes of the of the entities in the training data.  I can't share my specific data and use case, but you can find a similar one here in the parsing of resumes: https://dataturks.com/blog/named-entity-recognition-in-resumes.php  The resume is broken into distinct "sections" denoted by linebreaks.  The model can pick up on this information when for example identifying the skill section of the resume.

When outputting CSV format I agree that you would need to remove linebreaks, however I'd request that when outputting JSON the linebreaks be preserved.

BrambleXu:
Hi, @Zerthick. We have a discussion about your request. As for the conclusion, we won't support rendering line breaks for sequence labeling task and won't support exporting JSON file with linebreaks for sequence labeling task. This will cause the indent problem, which has a technical difficulty. 

Line breaks information might be important for document classification but not for NER task. And for NER task, there is no need to import the whole document as one text. We hope you can preprocess the text and split it to many lines, which not contain line breaks.

phoerious:
Not having line breaks makes the sequence labelling interface entirely useless for anything *but* NER. There should definitely a mode that renders line breaks. Right now I have to change the source code to support other use cases. Font sizes and line spacing are also way too large for anything that's more than a single sentence.

icoxfog417:
We mainly develop sequence labeling interface to do NER. Would you please show the use cases except NER/POS tagging? We want to know these.

Anyway, supporting linebreak (newline) requires careful consideration (refer: [support on Prodigy](https://support.prodi.gy/t/lengthy-newlines-etc-rendering-oddly-in-ui/557)).

phoerious:
That's any case of sequence labelling where you want to classify parts of a document, not parts of a sentence (extraction of headlines, quotations, paraphrases, style changes, paragraph-level sentiment, etc.).

shkarupa-alex:
+1 to support line breaks in sequence classification.
It would be very helpful in "sentence boundary detection" datasets.

PR #37, opened 2018-12-05
Enhancement/linebreak

BrambleXu:
iss27: render linebreaks in document classification and seq2seq tasks, and save the style to css rather than html

enhancement: add button to jump between edit data page and annotation page, and only show the edit button for admin user


Hironsan:
Thanks!

PR #39, opened 2018-12-05
Update README.txt to with a description of the new feature

PanBartosz:


BrambleXu:
@PanBartosz Thanks for your update. But the json feature only support import data now. We also want to support json format export feature in future. After we finish it, we will update the json part in README.

Issue #16, opened 2018-10-15
Don't track app database in git


jamesmf:
### Feature Request
It would be nice to not have the db tracked in git

### Describe the problem
The fact that the application database is tracked in git is troublesome if you ever want to interact with the repo. For example if a new feature gets implemented and I want to sync to the latest commit on master, I'd need to jump through hoops in order to pull down those changes without merge conflicts in the db.

This could also slow down development in forks (you'd need to reset the db every time you want to PR back into this repo).

PR #41, opened 2018-12-07
Not track app database in git

BrambleXu:
#16 
1. delete db and ignore db
2. update migration command in README


Hironsan:
Thanks!

Issue #43, opened 2018-12-13
Breaking line problem on sequence labeling task when selected text is too long.


Hironsan:
# System information
- Any OS

# Describe the problem

The following problem was reported by a user:

> I've accidentally selected almost entire text and annotated it. Now I cannot delete this annotation (pictures below). What could I do? Even if I knew how to redo it (the Ctrl+z does not work) I would not do it since I begun to annotate later texts. What should I do?

![image](https://user-images.githubusercontent.com/6737785/49917295-f9fcef80-fee1-11e8-8a61-4d87cd3693ee.png)

![image](https://user-images.githubusercontent.com/6737785/49917321-126d0a00-fee2-11e8-8273-071672fc6777.png)

# Solution
A quick-fix solution is to modify annotation.css as follows:

```css
.content .tag {
  font-size: 16pt;
  height: 1.5em;
  padding-left: .5em;
  padding-right: .5em;
  white-space: unset;  /* Add */
  height: auto;        /* Add */
}
```

But I think we will check the problem carefully.

BrambleXu:
Sure, I will reimplement this bug locally and then we can have a talk.

BrambleXu:
Beside this issue, we also solve the delete button margin problem in the label page.

Before
![image](https://user-images.githubusercontent.com/10768193/49921442-e0af6f80-fef0-11e8-9ce2-5b15ab2da789.png)

After
![image](https://user-images.githubusercontent.com/10768193/49921529-2a985580-fef1-11e8-8f9b-74e1eca116be.png)

PR #44, opened 2018-12-13
solve issue #43 for long selected text and delete button margin

BrambleXu:
#43 
iss43: add tagged-text class in js and css
fixed: delete button margin


BrambleXu:
Yes, you are right! Working on it.

Hironsan:
Thanks!

Issue #45, opened 2018-12-18
Feature Request: Add document metadata


serzh:
### Describe the problem
It would be nice if it possible to import some examples metadata while importing the dataset. For now it is possible only to import the text, but in my expirience I rarely need only text. I often need some external id, source, etc to be passed into the annotation tool along the text. And it would be nice to have it within my exported dataset.

BrambleXu:
Please give us more detail of your problem.
* Which task you need to import your metadata? Document classification? Sequence labeling? Sequence to sequence? Or all of them.
* If you often import external information, what kind of data format you use? Json?
* In the export data, take classification task as an example, you can find the document id, text, label, username. Is this not enough for your task?

serzh:
Sure, thanks for rapid feedback!

My task now is to provide annotations for several steps in the ML pipeline for information extraction. But it is not limited to my setup. I will try to describe more precisely when this problem arises.

Let's imagine you have a not annotated dataset, which contains some metadata for each example: id, source, date for creation, user that grabs it, etc. Now you want to use `doccano` to annotate it. To be able to match the initial dataset with annotated you need, at least, some `external_id` that is imported with your dataset into the `doccano`. 

I see two usages for doccano:
* source of truth for datasets and annotation tool
* annotation tool

And for both of them you need to have some metadata to be stored with your examples. 
It does not matter what type of task you are accomplishing and what is the format of you dataset. I see this as a generally usefull feature.

serzh:
It is not well documented but Explosion's annotation tool Prodigy does have this feature https://support.prodi.gy/t/how-to-incorporate-document-metadata/296/2

BrambleXu:
Thanks for your supplement. 
Your request does make sense. We do not consider such metadata like data source, user who grabs and so on. But doccano is a light annotation tool, it would be a nightmare to consider all kind of metadata. 

However, as you said, if there is at least the `external_id` with imported file and export file, you can align the data by yourself after you get the annotated file. It is a good idea. We will consider your request.

serzh:
>  user who grabs and so on. But doccano is a light annotation tool
We can take an approach used by Prodigy tool: just store all *other* fields into the json column in database at import phase and merge it on the export phase.

BrambleXu:
If we consider the `context` (many kinds of metadata), it may change the model classes a lot in `models.py`. I use 'light' to means a minimal modification to model classes and a few of fields in the database. Either way, I will talk with @Hironsan and decide which approach to use.

[...]

PR #55, opened 2018-12-27
Feature metadata

serzh:
This PR introducec a concept of metadata: any additional keys or columns in the imported file that are preserved and exported in the metadata key or column on export. The main use-case is to be able to match exported data with your existing corpus by providing `external_id` in the imported file, although there are other use-cases. Corresponding issue #45 .

Here are some comments about this PR:
1. **CSV vs TXT**
It is not clear for me why we have csv format called a txt format. It is clear that txt format that is described on the "Import" page is a subset of the csv. Why don't call it just CSV? That is what I have done. And to preserver previous behaviour I have made the CSV header and text column in it required, unless it is one-column csv (previous txt format). If it is one-column csv it is assumed that the only column is the text column.

2. **Importing big files**
In the previous implementation the whole file was read into memory on the import phase. That may cause memory problems with really big files. To mitigate this issue I have made processing of this files to use generators and to use batched creation of the `Document` rows in database. I have made batch size to be configured by the `IMPORT_BATCH_SIZE` setting.

3. **Error handling** 
There was several places in the code where caught exceptions was not logged anywhere. To fix this I have added logging of them. Also I have added `django.contrib.messages` facilities to the `admin_base` template to show application level errors. For now it is used only for import feature, but I think that there are many other places where it will be appropriate.

4. **`Project` polymorphism based on its type** (not done)
We have 3 project types: document classification, sequence labeling and seq2seq. All of this types are represented as a single `Project` class with lots of `if-else`s on project type. It will be more better to use inheritance and `proxy` model to represent these polymorphic relations. It will be easier to extend, write and reason about.


BrambleXu:
Thanks for your great work! 
Due to the year-end holiday, it might be a while to review this PR.

serzh:
@BrambleXu 
Regarding your first comment. This is regular CSV quotation: if csv value contains a double-quote, then all double-quotes are doubled and the whole value is enclosed with double-quotes. If you will read this file using any csv reader it will be read just fine.
Here is excerpt from [RFC 4180 Common Format and MIME Type for CSV Files](https://tools.ietf.org/html/rfc4180):
>>>
    7.  If double-quotes are used to enclose fields, then a double-quote
        appearing inside a field must be escaped by preceding it with
        another double quote.  For example:

        "aaa","b""bb","ccc"

serzh:
@BrambleXu Regardless the refactoring. Sure, I can open an issue and start working on the refactoring.

BrambleXu:
@serzh Thanks for your explanation. As you said, double-quotes can be read normally by CSV reader. 

Regardless of the refactoring, we need advice and feedback from @Hironsan. He is the main contributor after all, and he has a clear vision for doccano.

BrambleXu:
@serzh I talked with Hironsan,  we believe that you can do a great job. Please open an issue and refactor the models. 

This model graph is drawn by pygraphviz. Hope this is helpful.
image

PR #68, opened 2019-01-23
Feature/Add Dockerfile

c-w:
As per https://github.com/chakki-works/doccano/projects/2#card-16406087, this change adds a Dockerfile for the project.

The Python distribution used by the image can be specified via a build-time argument. This enables us in the future to for example run the tests on a wide variety of Python versions.

In the future, the Dockerfile could be extended to for example leverage multi-stage builds. In this case, the first stage could run tests, linter, and so forth and then create a wheel or similar artifact via setuptools. A second stage in the build can then only include a minimal set of dependencies to install and run the artifact (e.g. using python-slim as a base image).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+*
+!app/
+!data/
+!tests/
+!requirements.txt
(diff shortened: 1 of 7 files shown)


Hironsan:
Thanks for your contribution!

Issue #61, opened 2019-01-07
Unable to annotate any document with more than 26 labels


dveselov:
Hi,
First of all, I must say that doccano is pretty good.
But I come with problem: doccano becomes unusable, when I need to annotate any document with more than 26 labels (example task: classification of hand-written country names: `englind -> england`, `usa -> united_states_of_america` and so on), because label can't be created without shortcut, label shortcut can't be reused and shortcuts map contains only lowercased ascii characters.

Thanks.

dveselov:
As workaround, I've replaced `strings.ascii_lowercase` with `strings.printable` in label shortcuts choices: https://github.com/bureaucratic-labs/doccano/commit/39cc965b3618ed4fb39ab50b1fec140789e38cc2 & https://github.com/bureaucratic-labs/doccano/commit/859018bf0527dd087deec0709312e0b7755d08b4
Everything works fine for me, if this change is acceptable I can create PR (/cc @BrambleXu).

BrambleXu:
Hi @dveselov, I am glad you work this out. 

I test your code locally. I can add labels only by clicking, but can not use the short cut keys to add `@` and `A`. Do you also only add these label by clicking? If so, this is not acceptable for a PR. If not, please make the changes in your code. 

![image](https://user-images.githubusercontent.com/10768193/50883092-56dcbe80-142a-11e9-940a-fce459814277.png)

If you add these labels only by clicking, there is no need to increase the short cut keys. You can make the short keys optional instead. 

Find `class Label` in `models.py`, modify it as below:
```
shortcut = models.CharField(max_length=10, choices=KEY_CHOICES, blank=True)
class Meta:
    unique_together = (
        ('project', 'text'),
    )
```
Then run below command in command line.
```
python manage.py makemigrations
python manage.py migrate 
```
If you have done these steps, the short keys become optional. You can add unlimited labels withoud short cut keys.

![image](https://user-images.githubusercontent.com/10768193/50884630-c81e7080-142e-11e9-9006-b1ba11b34b7c.png)

Hironsan:
I think making short cut keys optional is a good idea. But at the same time, we should consider increasing the number of short cut keys. For example, we can increase them by combining ASCII characters and ctrl or alt keys.

aribornstein:
I am having this issue as well having more shortcut keys would be great but I can imagine a situation i.e co-reference taggings where I have hundreds of mention tags in which case there won't be enough shortcut keys. P.S this is an amazing project :)

PR #73, opened 2019-01-25
Enhancement/shortcuts 

BrambleXu:
This PR build a more powerful and user-friendly shortcut key setting. Major enhancement:

* Provide `ctrl`, `shift` key to create more shortcuts. So there are four combinations of shortcut: `a`, `ctrl a`, `shift a`, `ctrl shift a`. There are totally 4*26=104 shortcut keys for choose. I think this can cover most of the use-cases. 
* Make shortcut optional If user deals with more than 104 labels. So a user does not need to assign a shortcut to a label. 

Annotating Usage:
* For the label with a shortcut key, the user can annotate by shortcut key or clicking the label.
* For the label without a shortcut key, the user can annotate only by clicking the label.

Page of creating label:
image

Page of annotation:
image

Some implementation explanation for review easily.

forum.css:
* Correct typo

utils.py:
* Generate KEY_CHOICES

models.py:
* Make shortcut field optional and allow null=True

test_model.py:
* Create new test case for null shortcut 

label.js: 
* I use `selectedKey` to represent the alphabet and use `checkedKey` to represent the checktbox value. 
* `shortcutKey` is the final shortcut key add to the model.
* `combineKeys` is a Vue compeute and is used to combine `selectedKey` and `checkedKey`.
* `saveKeys` is used to save null to DB if the shortcutKey is null

filter.js:
* Show simplified shortcut key

mixin.js:
* Add function to show simplified shortcut key for document_classification.html & sequence_labeling.html

label.html:
* Add required and optional suggestion
* Add checkbox

document_classification.html & sequence_labeling.html:
* Change code for vue-shortkey format
* Show simplified shortcut key

### TODO

- [x] UI improvement: Show shortcut key more beautifully
- [x] Refactor shortcut field and front js to accept null directly
- [x] Rewrite all code for clean commit log

Corresponding issue #61


BrambleXu:
When calling `label.shortcut`, `CharNullField` class will convert NULL (stored in DB) to empty string `''`. So if we set many labels without shortcut key, `label.shortcut` will return these label as empty string, which is the reason why `test_models.py` is failed. 

As I explained above, if we make shortcut key optional, we should not set the shortcut key unique. Because the empty string for no-shortcut-labels can not pass the uniqueness check. 

I think we should delete the `test_shortcut_uniqueness` in `test_models.py`. What do you think? @Hironsan 
If it is ok, I will update a new commit.

Hironsan:
Due to bulma's default css(`code` tag), shortcut text is colored by red.

image

app/server/templates/admin/label.html
```diff
@@ -36,42 +35,61 @@
             <a class="tag is-medium" v-bind:style="{ color: textColor, backgroundColor: backgroundColor }">
               [[ labelText ]]
             </a>
-            <span class="tag is-medium">[[ selectedShortkey ]]</span>
+            <span class="tag is-medium"><code>[[ previewKeys ]]</code></span>
```
Hironsan: I think this line is somewhat redundant.
In label.js, `previewKeys` is almost equivalent to `simpleShortcut` filter.
So I think it is better to use `[[ combineKeys | simpleShortcut ]]` instead of using `previewKeys`.

Hironsan:
You can use `<code>` element but I think `<kbd>` is another choice.
Please consider it.

* [The Keyboard Input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd)

app/server/templates/admin/label.html
BrambleXu: Thanks, this can be simplified!

app/server/models.py
```diff
@@ -1,11 +1,10 @@
 import json
-import string
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.urls import reverse
 from django.contrib.auth.models import User
 from django.contrib.staticfiles.storage import staticfiles_storage
-
+from .utils import CharNullField, get_key_choices
```
Hironsan: Blank two lines.

Hironsan:
We wish we could define unique constraints in this way:

* [UniqueConstraint](https://docs.djangoproject.com/en/dev/ref/models/constraints/#uniqueconstraint)

Available from Django 2.2.

yangliu2:
With the latest version, there is a "shift" modifier for shortcut keys. However it give the error "You cannot use same label name or shortcut key." if you have the regular key taken.

BrambleXu:
@yangliu2 
Thanks for the feedback. Please raise an issue and give us the problem detail you met.

PR #75, opened 2019-01-25
Feature/Add social login via Github and Active Directory

c-w:
Currently the workflow for adding new users to a doccano instance requires an administrator to create accounts using the Django admin tool. This puts a burden on the admin. It also puts a burden on the user since they must maintain a dedicated account and password to access this application. See also the discussion on [card-16405801](https://github.com/chakki-works/doccano/projects/2#card-16405801) around user management.

To make it easier for annotators to do work on doccano, I suggest to add social authentication to the application. This means that a user can sign up with the social provider and get invited to work on annotation projects using that social identity.

As a first step, this pull request implements Github and Azure Active Directory authentication but other social identities should be easily added in the future. This pull request does not solve the problem of having an easy creation of admin users who can create new projects, but at least it makes it simple for annotators to join and work on a project.

Screenshot of the updated login page:

[Login page screenshot](https://user-images.githubusercontent.com/1086421/51774269-b17b4900-20bf-11e9-96d1-09625ffd19e4.png)

### DIFF
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -49,7 +49,7 @@ INSTALLED_APPS = [
     'server.apps.ServerConfig',
     'widget_tweaks',
     'rest_framework',
-    'django_filters'
+    'django_filters',
 ]
 
 MIDDLEWARE = [
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -50,6 +50,7 @@ INSTALLED_APPS = [
     'widget_tweaks',
     'rest_framework',
     'django_filters',
+    'social_django',
 ]
 
 MIDDLEWARE = [
@@ -60,6 +61,7 @@ MIDDLEWARE = [
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    'social_django.middleware.SocialAuthExceptionMiddleware',
 ]
 
 ROOT_URLCONF = 'app.urls'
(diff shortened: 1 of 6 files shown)


Hironsan:
Nice feature!
But I have a question.

In `urls.py`, is the following url right?

* https://github.com/chakki-works/doccano/pull/75/files#diff-3dc4970f50a896f3b3911da07a463c47L25

I tried to login vie GitHub OAuth, then I had an Authentication Failed error:
image

According to the [Python Social Auth documentation](https://python-social-auth-docs.readthedocs.io/en/latest/developer_intro.html), I changed the url from above to below. It works fine:

```python
# path('social/', include('social_django.urls')),
path(r'', include('social_django.urls', namespace='social')),
```

Do we need any additional settings?

c-w:
@Hironsan Unlike the social_django documentation, I've used `path('social/', include('social_django.urls')),` in this pull request to ensure that none of the included URLs from social_django will ever clash with URLs defined in the main application since setting the first argument to `social/` means that all the social_django URLs will be prefixed with that token.

So for example the Github complete URL will now be `/social/complete/github` instead of `/complete/github`. This simply requires the callback URL in the Github OAuth application to be configured slightly differently, to include the `/social` prefix. See the screenshot below for an example:

[Screenshot of Github OAuth application registration](https://user-images.githubusercontent.com/1086421/51810595-dbc33700-2276-11e9-8794-6f5797b4f559.png)

I believe that having the `social/` prefix has benefits since it means that developers of doccano don't have to be aware of all the potential URLs in the social_django project and we don't get odd bugs in the future of routes overriding each other.

Could you try with the branch's code and change the Github callback URL in your OAuth application registration to include the `social/` prefix and confirm that the feature then works for you? Thanks in advance!

Hironsan:
> I believe that having the social/ prefix has benefits since it means that developers of doccano don't have to be aware of all the potential URLs in the social_django project and we don't get odd bugs in the future of routes overriding each other.

I agree with you.

It works fine. Thank you!

Hironsan:
Simple documentation:
* https://github.com/chakki-works/doccano/wiki/How-to-use-OAuth

PR #76, opened 2019-01-25
Feature/Make client-side analytics configurable

c-w:
The doccano application currently contains Google Analytics with a hard-coded tracking ID. This means that every deployment of doccano will send analytics to the same account. In order to support new deployments of doccano bringing their own analytics, this change makes the Google Analytics tracking ID configurable via a setting.

Additionally, the Google Analytics Javascript will now only be included in the page if the tracking ID is present which means that deployments can also opt-out of Google Analytics altogether.

For users who're looking for an alternative to Google Analytics, this change also adds an option to use [Azure Application Insights](https://docs.microsoft.com/en-us/azure/azure-monitor/app/javascript) for client-side monitoring.


Hironsan:
> The doccano application currently contains Google Analytics with a hard-coded tracking ID.

To tell the truth, I deliberately embedded the hard-coded tracking ID. Because I wanted to know how many people are using doccano. Through Google Analytics, I can know the number and it motivated me a lot.

image

But I can understand what you say. This feature is useful for people who want to set their own ID.

Hmmm...

Is there any way to know how many users using doccano but at the same time the user can set their own ID?

Hironsan:
No problem. I really appreciate your help.

c-w:
Wow, great to see the usage. In any case, even after this merge, the tracking ID still defaults to yours; the user actively must set the environment variable to an empty string to remove the tracking (like `GOOGLE_TRACKING_ID=""`). Hope that helps.

PR #80, opened 2019-01-28
Enhancement/Auto-generate bundle files

c-w:
I noticed that the webpack outputs are currently checked into the repository. This is nice since it enables a simple deployment model with git push to Heroku. However, it also somewhat pollutes the repository and diffs since the bundle files are auto-generated; so it would be nice to remove these auto-generated files from the repo. Additionally, removing the generated files from the project means that the documentation now must contain instructions on how to build the frontend which helps with onboarding.

Given that we recently introduced a Dockerfile for the application, I suggest to move the Heroku deployment to use the Docker image and create the webpack assets as part of the Docker build. In this way we get the best of all the worlds: a simple setup for deployment and development (just a docker build and docker run) and no generated files checked into the repository.

This change will also make it easier to move to a fully integrated solution such as [Django Webpack Loader](https://github.com/owais/django-webpack-loader) in the future.


c-w:
Thanks for the merge!

PR #92, opened 2019-02-18
Group requirements

### LINKED ISSUE #90: Grouping requirements
[requirements.txt](https://github.com/tehamalab/doccano/blob/master/requirements.txt) seems to contain both base, development and deployment dependancies. This may lead users to install unnecessary packages on some cases, for example package like `flake8` may be unnecessay for production deployment  while `gunicorn` may be unnecesary for development installation and if someone wants to deploy doccano using a different application server like [uWSGI](https://uwsgi-docs.readthedocs.io/en/latest/).

I think it would be better to have separate requirements files for different use cases, for example

- `requirements.txt` for basic requirements
- `requirements_dev.txt` for development requirements
- `requirements_<gunicorn/something>.txt` for deployment or platform specific requirements

I have done that at https://github.com/tehamalab/doccano/tree/group-requirements . If it sounds good  I can submit a pull request

Additionally, why `tornado` is among the dependancies?

Thank you very much building doccano.

machakux:
closes #90


Hironsan:
I think Pipfile is a better option.

PR #99, opened 2019-02-27
revise the offset calculation for sequence labeling

LiyuanLucasLiu:
The current implementations relies on the js script (`https://github.com/chakki-works/doccano/blob/master/app/server/static/js/sequence_labeling.js` line44-45, line51-52) to calculate the start & end offset for annotated entities. Due to the behavior of the `length` function in javascript, these offsets are based on the number of utf-16 units (e.g., some emoji would have a length of 2). This PR fixes this issue by revising the annotation offsets. 

PS: Thanks for releasing the project :- ), love it

### DIFF
--- a/app/server/models.py
+++ b/app/server/models.py
@@ -154,7 +154,7 @@ class Document(models.Model):
                 start_mapping[cursor] = ind
                 cursor += len(text[ind].encode('utf-16-le')) // 2
                 end_mapping[cursor] = ind
-            for a in annotations
+            for a in annotations:
                 assert(a.start_offset in start_mapping)
                 a.start_offset = start_mapping[a.start_offset]
                 assert(a.end_offset in end_mapping)
--- a/app/server/models.py
+++ b/app/server/models.py
@@ -148,7 +148,18 @@ class Document(models.Model):
         if self.project.is_type_of(Project.DOCUMENT_CLASSIFICATION):
             return self.doc_annotations.all()
         elif self.project.is_type_of(Project.SEQUENCE_LABELING):
-            return self.seq_annotations.all()
+            annotations = self.seq_annotations.all()
+            start_mapping, end_mapping, cursor = dict(), dict(), 0
+            for ind in range(len(self.text)):
+                start_mapping[cursor] = ind
+                cursor += len(text[ind].encode('utf-16-le')) // 2
+                end_mapping[cursor] = ind
+            for a in annotations
+                assert(a.start_offset in start_mapping)
+                a.start_offset = start_mapping[a.start_offset]
+                assert(a.end_offset in end_mapping)
+                a.end_offset = end_mapping[a.end_offset]
+            return annotations
         elif self.project.is_type_of(Project.Seq2seq):
             return self.seq2seq_annotations.all()
 
--- a/app/server/models.py
+++ b/app/server/models.py
@@ -149,16 +149,16 @@ class Document(models.Model):
             return self.doc_annotations.all()
         elif self.project.is_type_of(Project.SEQUENCE_LABELING):
             annotations = self.seq_annotations.all()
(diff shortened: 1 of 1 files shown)


Hironsan:
Hi @LiyuanLucasLiu 

Thanks for your PR. I just understood the problem.

But I think it is better to convert a character in JavaScript to a unicode by using some method like [normalize method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize). Because Python represents a character as a unicode so we can maybe solve the problem simpler.

Is there any idea?

LiyuanLucasLiu:
Actually I think it would be better to directly insert the annotations to the original text (e.g., using <PER>Obama</PER>) instead of recording the offset.

Hironsan:
Like following?

```python
class SequenceAnnotation(Annotation):
    document = models.ForeignKey(...)
    label = models.ForeignKey(Label, on_delete=models.CASCADE)
    start_offset = models.IntegerField()
    end_offset = models.IntegerField()
    text = models.CharField()  # add
```

Or, remove offset?

```python
class SequenceAnnotation(Annotation):
    document = models.ForeignKey(...)
    label = models.ForeignKey(Label, on_delete=models.CASCADE)
    text = models.CharField()
```

Are there any merits to insert text directly into a model?

LiyuanLucasLiu:
I think offsets is kind-of ambiguous (depend on the way to calculate the length), while the inline annotations are self-explanatory :-)

Hironsan:
Do you mean the model has an entire text that has inline annotations instead of offsets?

For example, we save the follwing text into the model:

```
<PER>Barack Hussein Obama II</PER> (born <DATE>August 4, 1961</DATE>) is an <OTHER>American</OTHER> attorney and politician.
```

Instead of:

```
{
    id: 16,
    label: 1,
    start_offset: 0,
    end_offset: 23,
},
{
    id: 17,
    label: 5,
    start_offset: 30,
    end_offset: 44,
}
...
```

LiyuanLucasLiu:
Yes, exactly :-)

BrambleXu:
@LiyuanLucasLiu Your final goal is to get the XML format as export. Do I understand right? If so, you can translate the export file from JSON format to XML format, because the JSON file already contains the information you need.

LiyuanLucasLiu:
@BrambleXu My final goal is to solve the disparity between utf-16 and unicode. This PR proposes a change to calculate the unicode index based on the utf-16 index.

LiyuanLucasLiu:
And i don't think normalization is a good idea, because it would change the original data.

BrambleXu:
If you only implement for the utf-16, what if someone uses utf-32 encoding? This uncertainty will bring some bug in the future. Convert these to a normalized formal is a good choice. Even if the data changes, I don't think this would affect the training process due to two encoding is the same meaning(the same emoji character).

LiyuanLucasLiu:
@BrambleXu Emmm... I think you mis-understood my point. The current problem is: no matter what kind of encoding is used, the annotations for sequence labeling would be recorded by a js script. Unfortunately, in javascript, the default `length` function is not to calculate the number of unicode, but to calculate the number of utf-16 units. So, in order to be compatible with the python script which uses the unicode encoding, i try to map the utf-16 index to the unicode index in this pr. In other words, this pr tries to fix the bug that current implementation cannot handle emoji properly. 

I agree to use unicode as the text encoding (and originally i thought by normalization you mean text normalization).

BrambleXu:
@LiyuanLucasLiu Thanks for the explanation. I am clear now.

BrambleXu:
@LiyuanLucasLiu Could you provide some data samples you used? I want to what your exact use case for annotating emoji. The data samples will be used for testing.

LiyuanLucasLiu:
I'm creating a twitter NER benchmark. Typically any tweets with emoji would cause this problem.

PR #103, opened 2019-03-01
Enhancement/Add dotenv support

c-w:
Resolves https://github.com/chakki-works/doccano/issues/94

Previously we were able to configure some aspects of the project via environment variables. This pull request adds the option to configure these settings in a `.env` file placed in the `app/.env` folder.

The implementation uses [environs](https://pypi.org/project/environs/) which wraps [python-dotenv](https://pypi.org/project/python-dotenv/) and adds a bunch of nice convenience methods such as automatic type casting and validation of environment variables (e.g. to convert from string to list or from string to boolean).


Hironsan:
Thanks!

PR #117, opened 2019-03-22
Enhancement/Integrate Webpack and Django

c-w:
This is a follow-up to https://github.com/chakki-works/doccano/pull/80.

This pull request more tightly integrates Django asset serving with the Webpack compiler. Specifically, after this change, it will no longer be required to manually compile the project's Javascript to integrate it with the Django server. Instead the Javascript will be compiled and reloaded on the fly automatically.

In a development scenario (here emulated by running the server as described in the README), the assets get linked via the hot reloading Webpack development server:

Screenshot showing development asset loading

This enables a number of enhancements to the developer experience, including not having to manually run the Webpack build and also being able to debug Javascript before it went through the build process:

Screenshot showing debugging of Javascript before compilation

In a production scenario (here emulated by running the server via Docker), the assets get directly linked from the compiled static bundles folder as before:

Screenshot showing production asset loading

### DIFF
--- a/app/server/webpack.config.js
+++ b/app/server/webpack.config.js
@@ -1,8 +1,12 @@
 const process = require('process');
+const BundleTracker = require('webpack-bundle-tracker');
 const VueLoaderPlugin = require('vue-loader/lib/plugin')
 
+const devMode = process.env.DEBUG !== 'False';
+const hotReload = process.env.HOT_RELOAD === '1';
+
 module.exports = {
-    mode: process.env.DEBUG === 'False' ? 'production' : 'development',
+    mode: devMode ? 'development' : 'production',
     entry: {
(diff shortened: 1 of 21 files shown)


app/server/webpack.config.js
```diff
@@ -18,9 +22,16 @@ module.exports = {
         'dataset': './static/js/dataset.js',
     },
     output: {
+        publicPath: hotReload ? 'http://localhost:8080/' : '',
```
Hironsan: Probably http://localhost:8000 ?

Hironsan:
Thanks. But I faced a problem.

When I set DEBUG=False in environment variable and run development server, the server returned 500 server error in http://127.0.0.1:8000/projects/.

Is there any way to solve the problem?

Hironsan:
Thanks!

PR #118, opened 2019-03-22
Enhancement/Turn dockerfile into a multi-stage build

c-w:
Using a multi-stage build reduces the image size from 1.23GB to 331MB which means that any node running doccano can now start-up faster.

Additionally, the multi-stage build means that we no longer include in the production image development tools like node (now excluded since it's only installed in the builder layer) or gcc (now excluded since the production layer is based on python-slim) which reduces the attack surface of the image.


Hironsan:
Awesome!

PR #148, opened 2019-04-10
Enhancement/Refactor projects page to Vue single file component

c-w:
Currently Django and Vue are somewhat coupled in that the project intermixes templating constructs from both eco-systems.

Examples of this in `projects.html` include:
- Checking if the user is a superuser to modify the Vue template.
- Using Django filters to inject the capitalized username.

This makes it somewhat hard to follow the Vue code and also makes it more difficult to modularize the Vue components and re-use them.

As such, this change improves the separation of concerns by having the Django template only scafold the root element for Vue and serialize any variables that are required from the Django context to Javascript so that Vue can access them. The rest of the templating is then moved into a Vue single file component. This will make it easier for the developer to understand the Vue components and make it very clear what is rendered by Django and what is rendered by Vue.

In order to keep the changes minimal and focused, for now, only the projects page is moved to the new single file component paradigm. In future changes, some of the other pages can also be converted by following the same approach as outlined in this change. In future changes, the projects component can also be split into more smaller components to even better leverage the Vue single file component mechanism.

The change also adds eslint for single-file components. The linter acts on Vue components in both `.js` and `.vue` files. To keep the changes minimal and targeted, for now the linter uses a very permissive rule-set which essentially means that the linter is mostly disabled. As a separate change (e.g. see [this branch](https://github.com/CatalystCode/doccano/compare/enhancement/refactor-projects-page-to-vue-single-file-component...CatalystCode:enhancement/fix-vue-eslint-errors)), we should fix all the lint issues and tighten the rule-set to `plugin:vue/recommended`.


app/server/templates/projects.html
```diff
@@ -3,151 +3,10 @@
...
-      </div>
-    </div>
-  </section>
-
-</div>
+<div id="projects_root"></div>
 {% endblock %}
 {% block footer %}
+{{ user.get_username | json_script:"django.username" }}
+{{ user.is_superuser | json_script:"django.is_superuser" }}
```
Hironsan: I think Django and Vue are somewhat coupled in these lines.
Is the Me API I submitted in https://github.com/chakki-works/doccano/pull/151 useful for decoupling?

c-w: Yes! This would be a great next step refactor. We can either first merge the Me API and then I can update this PR or we can first merge the PR, then merge Me API and refactor in the future to further decrease the coupling.

Hironsan: OK, first merge the Me API and update this PR.

c-w: Excellent. Please ping this thread when the Me API is merged and I'll make the required changes.

Hironsan: Merged.

c-w: Refactored to use the /me API in https://github.com/chakki-works/doccano/pull/148/commits/72d91c6188f91e5bfe1a873431b8868353bd5677.

PR #151, opened 2019-04-11
Feature: Implement me API

Hironsan:
Currently, when checking whether the user is a superuser or not, or showing the username, it depends on the Django context.

This PR adds `me API`, which returns the user information. For using this API, we will be able to reduce the dependency.

Me API can return information like follows:

```
{
    "id": 1,
    "username": "hironsan",
    "first_name": "",
    "last_name": "",
    "email": "",
    "is_superuser": true
}
```

### DIFF
--- a/app/server/api.py
+++ b/app/server/api.py
@@ -14,11 +14,19 @@ from rest_framework_csv.renderers import CSVRenderer
 from .filters import DocumentFilter
 from .models import Project, Label, Document
 from .permissions import IsAdminUserAndWriteOnly, IsProjectUser, IsOwnAnnotation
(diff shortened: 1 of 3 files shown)


BrambleXu:
In such a case, the user authority only detects by "is_superuser" value. So there are mainly two [roles](https://github.com/chakki-works/doccano/issues/58#issuecomment-479380051),  Administrator and Labeler. Is there any plan for the Auditor role? Or we just set two roles right now.

Hironsan:
This PR focuses on decoupling, not role. it should be discussed in other issue. Please create corresponding issue.

PR #159, opened 2019-04-13
Enhancement/Remove Django template inheritance from annotation pages

c-w:
This pull request is a follow-up to https://github.com/chakki-works/doccano/pull/148 and moves the annotation pages to Vue single-file components. This change also fully decouples the Vue frontend for the annotation from the Django server by removing all usage of Django block-based template inheritance.

The main approach to removing the Django template inheritance is modeled after the article [Extending VueJS Components](https://vuejsdevelopers.com/2017/06/11/vue-js-extending-components/) and leverages the [Pug](https://github.com/pugjs/pug) pre-processor for Vue templates.

In the future, a separate refactor could further simplify the approach and remove Pug in favor of splitting the Vue components into smaller independent pieces and using slots for the content composition. This would be cleaner overall solution, but it would also be a much more invasive refactoring. In any case, I believe Pug is a good step in the right direction and will make these future refactorings easier since now all the templating code is in Vue. Having all the templating code in Vue also makes other efforts such as component re-use or testing much simpler. Alternatively, if the project decides that Pug offers value on its own due to its improved terseness over plain HTML templates, other Vue components could also be converted to Pug in the future to make the codebase more uniform.

The Pug approach demonstrated in this pull request can also be extended to the admin pages in Doccano, for example to replace the Django template inheritance on the upload and download screens, but I haven't tackled this yet.

Unfortunately the pull request diff is somewhat large, so please don't hesitate to reach out with questions or comments. However, most of the changes made were very mechanical and so the overall pull request should be low risk. See below for an outline of the refactoring approach. Steps 1-3 below are in commit https://github.com/chakki-works/doccano/commit/d5bc9e428e5ca0ad4d302c22c2224752c144d9a4. Step 4 is commit https://github.com/chakki-works/doccano/commit/bcfaad4a63b7652ea3b1a747f3d41418668da120.

1. For each webpack bundle, copy the source Javascript file into the `<script>` section of a new Vue single file component. Keep all the code associated with creating a new Vue instance in the webpack bundle Javascript file and remove those lines from the Vue single file component so that the file only exports a pure Vue component without any side-effects.

2. For each Django base template (e.g. `annotation_base.html`), move all HTML associated with a Vue template into a Pug file. Convert all the HTML to the equivalent Pug code and replace each Django block with a Pug block.

3. For each Django leaf template (e.g. `document_classification.html`), move all HTML associated with a Vue template to the `<template>` section of the appropriate Vue single file component. If the Django template referenced a block, convert the HTML to Pug and extend the appropriate Pug base template.

4. All the Django leaf templates now only differ in the webpack bundle they include to scaffold the page, so delete them all in favor of a single base template which uses the Django context to configure which webpack bundle should be injected. Update all the views in Django to use the new base template and set the appropriate context variable for the webpack include.

In order to ensure that all Pug templates remain high quality going forward, I've also added a linter for Pug in commit https://github.com/chakki-works/doccano/commit/ab473e317bbc3db70ccccfcd343d5f0b9e9a7027.


app/server/static/js/annotator.vue
```diff
@@ -0,0 +1,168 @@
...
+        left = e.end_offset;
+      }
+      const l = this.makeLabel(left, this.text.length);
+      res.push(l);
+
+      return res;
+    },
+
+    chunksWithLabel() {
+      return this.chunks().filter(r => this.id2label[r.label]);
```
Hironsan: This line produces the following warn:

```
[Vue warn]: Error in render: "TypeError: this.chunks is not a function"

found in

---> <Annotator>
       <SequenceLabeling> at static/js/sequence_labeling.vue
         <Root>
```

By updating the code as follows, the warn will not occur.

```
return this.chunks.filter(r => this.id2label[r.label]);
```

Hironsan:
And It will occur something strange in a sequence labeling project:

image

app/server/static/js/annotator.vue
c-w: Good catch. Fixed in af26d38.

c-w:
@Hironsan Thanks for the initial review. Could you share the data file (or a subset thereof) that exhibits the strange behavior in the sequence labeling project?

Hironsan:
Sure! 
https://github.com/Hironsan/anago/blob/master/data/conll2003/en/ner/train.txt

c-w:
@Hironsan Thanks for providing the source file. I was able to reproduce the issue and fixed it in e5fc2db:

Screenshot showing fixed sequence labeling UI

c-w:
Based on the work in this pull request, I now also have a branch that moves the admin pages to Vue single-file components ([see diff](https://github.com/CatalystCode/doccano/compare/enhancement/remove-django-template-inheritance-from-annotation...CatalystCode:enhancement/remove-django-template-inheritance-from-admin)). I will submit these changes once this PR is fully discussed, reviewed and merged.

c-w:
@Hironsan Just following up on this pull request. Do you have any further questions or concerns about this change that I can help address?

c-w:
Thanks for the merge! I've now also submitted the follow-up PR which applies the same refactor to the admin pages.

PR #167, opened 2019-04-21
Feature: take database settings from env variables

dveselov:
Hi,
This patch allows to take database settings from `env` variables.
Its especially useful, for those who can't or didn't want to use default database (sqlite), e.g. for me :)

### DIFF
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -10,7 +10,7 @@ https://docs.djangoproject.com/en/2.0/ref/settings/
 Any setting that is configured via an environment variable may
 also be set in a `.env` file in the project base directory.
 """
-from os import path
+from os import path, environ
 
 import django_heroku
 import dj_database_url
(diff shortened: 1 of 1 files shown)


c-w:
This feature should already be supported by setting the `DATABASE_URL` environment variable. See the code here:

https://github.com/chakki-works/doccano/blob/ec7cf38017e75f8da844386c0d4ad2b5d3b907a0/app/app/settings.py#L197-L198

The feature is for example being used by the Azure deployment script to make Doccano store data in PostgreSQL ([code pointer 1](https://github.com/chakki-works/doccano/blob/ec7cf38017e75f8da844386c0d4ad2b5d3b907a0/azuredeploy.json#L264-L265), [code pointer 2](https://github.com/chakki-works/doccano/blob/ec7cf38017e75f8da844386c0d4ad2b5d3b907a0/azuredeploy.json#L100)).

dveselov:
Oh, thanks, `DATABASE_URL` seems to be exactly what I want, but without `ssl=True` hardcoded argument, since I'm running `doccano` inside docker with default postgres image.
Can we get rid of enabled-by-default ssl?

c-w:
Let me look into that and get back to you.

c-w:
@dveselov Take a look at https://github.com/chakki-works/doccano/pull/170. Once this PR is merged, the settings will honor the value of `sslmode` if it's present in `DATABASE_URL` which enables you to switch off SSL explicitly. For example, to do so for Postgres, you could set:

```sh
DATABASE_URL=pgsql://user:password@host:5432/database?sslmode=disable
```

dveselov:
Great, thank you.

PR #170, opened 2019-04-22
Enhancement/Enable non-SSL connections to the database

c-w:
Currently, whenever the database is changed from the default SQLite via the DATABASE_URL environment variable, the connection to the database is forced to be SSL. This is a great default for production use-cases. However, in some scenarios, users may not want to use SSL, e.g. when testing against a local MySQL or PostgreSQL database.

This change enables users to explicitly request a non-SSL connection to the database by passing the `?sslmode=disable` (or similar) query argument in the DATABASE_URL environment variable.

For backwards compatibility, if the `sslmode` is not set explicitly in the connection URL, the code still defaults to requiring SSL on the connection.

### DIFF
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -15,6 +15,7 @@ from os import path
 import django_heroku
 import dj_database_url
 from environs import Env
+from furl import furl
 
 
 # Build paths inside the project like this: path.join(BASE_DIR, ...)
@@ -195,7 +196,11 @@ LOGIN_REDIRECT_URL = '/projects/'
 LOGOUT_REDIRECT_URL = '/'
 
 # Change 'default' database configuration with $DATABASE_URL.
-DATABASES['default'].update(dj_database_url.config(conn_max_age=500, ssl_require=True))
+DATABASES['default'].update(dj_database_url.config(
+    env='DATABASE_URL',
+    conn_max_age=500,
+    ssl_require='sslmode' not in furl(env('DATABASE_URL', '')).args,
+))
 
 # Honor the 'X-Forwarded-Proto' header for request.is_secure()
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -198,7 +198,7 @@ LOGOUT_REDIRECT_URL = '/'
 # Change 'default' database configuration with $DATABASE_URL.
 DATABASES['default'].update(dj_database_url.config(
     env='DATABASE_URL',
-    conn_max_age=500,
+    conn_max_age=env.int('DATABASE_CONN_MAX_AGE', 500),
     ssl_require='sslmode' not in furl(env('DATABASE_URL', '')).args,
 ))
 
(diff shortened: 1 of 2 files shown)


dveselov:
Can this be merged, please?

c-w:
Tagging @Hironsan for visibility.

Hironsan:
Merged. Thanks!

deltatree:
New merge leads to an error: 
File "/usr/local/lib/python3.6/site-packages/django/db/backends/sqlite3/base.py", line 159, in get_new_connection


    conn = Database.connect(**conn_params)


TypeError: 'sslmode' is an invalid keyword argument for this function

dveselov:
@c-w looks like `dj-database-url` doesn't react to `sslmode` argument at all :(
I've got following `DATABASES` content with that URL - `postgresql://postgres:mysecretpassword@database/postgres`:
```python
{
    'default': {
        'NAME': 'postgres',
        'USER': 'postgres',
        'PASSWORD': 'mysecretpassword',
        'HOST': 'database',
        'PORT': '',
        'CONN_MAX_AGE': 600,
        'OPTIONS': {
            'sslmode': 'require'
        },
        'ENGINE': 'django.db.backends.postgresql_psycopg2'
    }
}
```

c-w:
@dveselov Could you please confirm that everything is set up correctly? See screenshot below for the database entry that I see when running with commit 6ee23ec.

Screenshot showing DATABASE_URL settings

c-w:
@deltatree Could you please let me know what environment settings you're using? With the default SQLite mode, I'm seeing correct behavior any errors and this was confirmed in the CI too.

Screenshot showing default SQLite database connection

c-w:
@deltatree I confirmed that this is an issue in [dj-database-url](https://pypi.org/project/dj-database-url/) which [always sets the sslmode even if the underlying engine doesn't support SSL](https://github.com/kennethreitz/dj-database-url/blob/v0.5.0/dj_database_url.py#L127-L128). I've implemented a work-around for this in https://github.com/chakki-works/doccano/pull/189.

I also tested the behavior with an older version of the code before the merge of this pull request and it seems to exhibit the same behavior so I believe this is not a regression but a newly discovered bug in its own right. Just to confirm, could you please share your environment settings so that I can properly reproduce? Thanks!

dveselov:
@c-w can be reproduced via
```bash
$ docker run -ti --rm -e "DATABASE_URL=postgresql://postgres:passwd@database/postgres?sslmode=disabled" chakkiworks/doccano:latest python app/manage.py shell

Python 3.6.8 (default, Mar 27 2019, 08:53:45) 
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from django.conf import settings
>>> settings.DATABASES['default']
{'sslmode': 'require'}
```

c-w:
@dveselov Thanks for providing the repro steps. I can confirm the behavior.

Looks like this is an issue in [django-heroku](https://pypi.org/project/django-heroku/) which seems to [always override sslmode=True](https://github.com/heroku/django-heroku/blob/v0.3.0/django_heroku/core.py#L68-L69).

I've pushed a work-around to https://github.com/chakki-works/doccano/pull/189 in commit e5c935a.

PR #173, opened 2019-04-24
Enhancement/Remove empty lines from upload

c-w:
As discussed in https://github.com/chakki-works/doccano/issues/168, having empty lines in the upload data files can cause user confusion since they lead to the entire upload being rejected.

As such, this change filters out empty lines in upload text and json files instead of failing the upload.

### DIFF
--- a/app/server/tests/data/example.empty.jsonl
+++ b/app/server/tests/data/example.empty.jsonl
@@ -1,0 +1,5 @@
+{"text": "example"}
+
+{"text": "example"}
+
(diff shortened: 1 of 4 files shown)


icoxfog417:
I think we should not "ignore" the blank line. It'll be good to provide the chance to notice that "your upload file is not fully registered to doccano, is it ok?".

c-w:
Thanks for the comment. Sounds good. Closing this PR.

PR #180, opened 2019-04-26
Enhancement/Organize webpack entrypoints

c-w:
Since the move to Vue single-file components (in https://github.com/chakki-works/doccano/pull/159 and https://github.com/chakki-works/doccano/pull/176), the number of webpack entrypoints has greatly increased and the files somewhat clutter up the frontend folder.

As such, to increase discoverability and organization of the code, this change moves all the webpack entrypoints into a single folder so that the files can more easily be ignored when not adding or changing an entrypoint.


Hironsan:
I have the same opinion that It's hard to understand the frontend folder structure. So I want to make the project structure easy to understand for front-end developers.

For that purpose, I think we should adopt the project structure that is often used by front-end developers like nuxt one:

```
nuxt-example
├── assets
├── components
├── layouts
├── middleware
├── node_modules
├── pages
├── plugins
├── static
└── store
```

Is there any ideas?

c-w:
Interesting suggestion. Let me take a closer look at what the folders mean according to the [docs](https://nuxtjs.org/guide/directory-structure/) and I'll make a new pull request.

PR #181, opened 2019-04-26
Enhancement/Move projects Vue template to pug

c-w:
All Vue templates in the code base except for the Projects component are written in pug, so this change makes the codebase consistent by also porting the Projects component template from html to pug. This is a pure refactor without any functional change.

### DIFF
--- a/app/server/static/js/projects.vue
+++ b/app/server/static/js/projects.vue
@@ -1,196 +1,110 @@
-<template>
-  <div v-cloak>
-    <section class="hero project-image">
-      <div class="container">
-        <div class="columns">
-          <div class="column is-10 is-offset-1">
-            <h1 class="title is-1 has-text-white">
-              Hello, {{ username | title }}.
-            </h1>
-            <h2 class="subtitle is-4 has-text-white">
-              I hope you are having a great day!
-            </h2>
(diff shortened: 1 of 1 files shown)


app/server/static/js/projects.vue
```diff
@@ -1,196 +1,110 @@
...
+                            img.project-thumbnail(
+                              v-bind:src="project.image"
+                              alt="Project thumbnail"
+                            )
+
+                          div.dataset-item__main.is-vertical
+                            div.dataset-item__main-title
+                              div.dataset-item__main-title-link.dataset-item__link
+                                a.has-text-black(v-bind:href="'/projects/' + project.id")
+                                | {{ project.name }}
```
Hironsan: This line generates the following code:

```html
<div class="dataset-item__main-title-link dataset-item__link">
    <a href="/projects/2" class="has-text-black"></a>
    NER
</div>
```

The link is not clickable.

Is the following code correct?

```pug
a.has-text-black(v-bind:href="'/projects/' + project.id") {{ project.name }}
```

c-w: Good catch. Fixed in d1fb9f4.

PR #185, opened 2019-04-30
Change ami for awsdeploy and update the SequenceAnnotation serializer

RafaelZarate:
This PR fixes an issue with a invalid AMI id in awsdeploy.yaml (switched it to an Ubuntu 18.04 AMI) and added a new field to SequenceAnnotationSerializer to have more explicit label data when exporting.

### DIFF
--- a/app/server/serializers.py
+++ b/app/server/serializers.py
@@ -137,7 +137,7 @@ class SequenceAnnotationSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = SequenceAnnotation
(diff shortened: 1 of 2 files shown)


BrambleXu:
It would be better to use one PR for just one problem. As for the AWS deployment, I think there is no need to raise a PR just for updating to Ubuntu 18.04. In addition to SequenceAnnotationSerializer, you should raise a new PR or a new issue first. [This](https://github.com/chakki-works/doccano/wiki/How-to-Contribute-to-Doccano-Project) might be helpful.

PR #189, opened 2019-05-08
Bugfix/Fix sslmode with custom sqlite path and on Heroku

c-w:
There is an issue in [dj-database-url](https://pypi.org/project/dj-database-url/) which [always sets the sslmode even if the underlying engine doesn't support SSL](https://github.com/kennethreitz/dj-database-url/blob/v0.5.0/dj_database_url.py#L127-L128). See also the [conversation](https://github.com/chakki-works/doccano/pull/170#issuecomment-490023067) with @deltatree on https://github.com/chakki-works/doccano/pull/170 for more information.

There also is an issue in [django-heroku](https://pypi.org/project/django-heroku/) which seems to always override sslmode=True. See also the [conversation](https://github.com/chakki-works/doccano/pull/170#issuecomment-490543552) with @dveselov on https://github.com/chakki-works/doccano/pull/170 for more information.

This change implements a work-around for both behaviors.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,6 @@ RUN pip install --no-cache-dir /deps/*.whl
 
 COPY --from=builder /doccano /doccano
 
-ENV IS_HEROKU="False"
 ENV DEBUG="True"
 ENV SECRET_KEY="change-me-in-production"
 ENV PORT="80"
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,6 +33,7 @@ RUN pip install --no-cache-dir /deps/*.whl
 
 COPY --from=builder /doccano /doccano
 
+ENV IS_HEROKU="False"
 ENV DEBUG="True"
 ENV SECRET_KEY="change-me-in-production"
 ENV PORT="80"
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -220,3 +220,7 @@ APPLICATION_INSIGHTS = {
 }
 
 django_heroku.settings(locals(), test_runner=False)
+
+# work-around for dj-database-url: explicitly disable ssl for sqlite
+if DATABASES['default'].get('ENGINE') == 'django.db.backends.sqlite3':
+    DATABASES['default'].get('OPTIONS', {}).pop('sslmode', None)
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -219,7 +219,9 @@ APPLICATION_INSIGHTS = {
     'ikey': AZURE_APPINSIGHTS_IKEY if AZURE_APPINSIGHTS_IKEY else None,
 }
 
-django_heroku.settings(locals(), test_runner=False)
+# work-around for django-heroku: don't overwrite sslmode outside of heroku
+if env.bool('IS_HEROKU', True):
+    django_heroku.settings(locals(), test_runner=False)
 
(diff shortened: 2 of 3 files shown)


icoxfog417:
Let me arrange the issue (it'll be a little bothering, but please forgive me because it is security matter).

**Problem**

User can not use non-SSL connection at the development environment.

**Cause**

* `dj_database_url` allow the user by connecting non-SSL by `ssl_require` parameter.
* But `django-heroku` always overwrite this (and other) setting (it is already reported at https://github.com/heroku/django-heroku/issues/10).

**Solution**

We can overwrite database setting after `django-heroku` ([example from](https://github.com/heroku/django-heroku/issues/10#issuecomment-376896459)).

```py
import django_heroku
import dj_database_url
django_heroku.settings(locals())
# override DATABASE_URL set by django_heroku because it forces SSL mode locally
ssl_require = os.environ['ENV'] != 'development'
locals()['DATABASES']['default'] = dj_database_url.config(
    conn_max_age=django_heroku.MAX_CONN_AGE, ssl_require=ssl_require)
```

c-w:
@icoxfog417 Thanks for the review. Calling `django_herok.settings(...)` earlier is a great idea! I've moved the call in 89372b6 which also lets us remove the `IS_HEROKU` flag.

As for using the `DEBUG` flag to disable SSL, why not trust the user's input? If the user does not set `?sslmode=...` in `DATABASE_URL`, we will default to having SSL required as before. However, if the user knows what they're doing and explicitly passes `sslmode=...` we should probably respect that explicit setting instead of requiring another environment variable to be set (e.g. see [principle of least astonishment](http://wiki.c2.com/?PrincipleOfLeastAstonishment)). Curious to hear your thoughts!

icoxfog417:
@c-w Thank you for your hard work! 

To use `DEBUG` is an alternative idea for `IS_HEROKU`. But we can overcome the problem just only change the position of `django_herok.settings`. It's great and environment variable is no more need.

Finally, let me confirm the code.

```py
ssl_require='sslmode' not in furl(env('DATABASE_URL', '')).args
```

* `True`: `ssl_mode` does not exist in the `DATABASE_URL`
* `False`: `ssl_mode` exsit in the `DATABASE_URL`
  * **If `ssl_mode` is `require`**

So please add test code for `sslmode=require` case.

c-w:
@icoxfog417 You're correct. if `ssl_require=True` is passed into the `dj_database_url.config` function, SSL will always be set, no matter what is the value `DATABASE_URL`. If `ssl_require=False`, whatever value is present in `DATABASE_URL` will be used, or SSL will be disabled if the URL does not set anything. This is why we do the check: if `DATABASE_URL` does not set an explicit `sslmode`, we default to enabling SSL for secure defaults and to be backwards compatible with earlier behavior. However, if `DATABASE_URL` contains an explicit `sslmode`, we set `ssl_require=False` such that the sslmode provided by the user will not be overwritten.

I've added the new test case in 756fff8.

icoxfog417:
Thank you for your fix! 
@Hironsan Please check & merge if it ok.

dveselov:
Are there any obstacles to merge it, guys?

icoxfog417:
Sorry for the late reply. We were confirming Heroku deploy (it 👍 ).

dveselov:
Everything works good, thanks!

PR #195, opened 2019-05-13
Enhancement/Go to annotation page

c-w:
Currently a user can go to the previous annotation item and the next annotation item. There are 5 annotation items per page so if the user wants to go to a item in the middle of a long project, they may have to do a lot of clicks.

In order to improve the user experience, this pull request implements a new navigation that lets the user quickly jump to a specific annotation page.

This feature was requested by a number of users, including @axelinternet and @bwkchee.

The animation below shows the behavior of the new navigation element. The animation demonstrates navigating to a particular page as well as protection against too low and too high inputs:

Animation showing the new go to page feature

Resolves https://github.com/chakki-works/doccano/issues/58
Resolves https://github.com/chakki-works/doccano/issues/84

### DIFF
--- a/app/server/package-lock.json
+++ b/app/server/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "^7.0.0"
+        "@babel/highlight": "7.0.0"
       }
     },
     "@babel/generator": {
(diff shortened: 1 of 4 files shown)


Hironsan:
I don't think users want a pagination feature. What they really want is a resume feature. With pagination feature, a user can go to a specific page that he/she wants to resume annotation. But it requires him/her to write the page number down.

I think there are two strategies to implement a resume feature:

* client-side strategy
* server-side strategy

In the client-side strategy, the application implicitly saves the page offset and use it to resume annotation. When the user moves to the annotation page, the application automatically loads the page offset and move to the page. Each time the user loads 5 annotation items, the application saves the page offset to some storage like localStorage. One of the advantages of this method is it can reuse this PR code. A disadvantage is a user cannot resume annotation when he/she use a different device from the one he/she has annotated.

In the server-side strategy, the application explicitly requires a user to check whether he has been completed annotation for a document or not. Once he finished annotation for a document, he needs to check "Checkbox". If he wants to resume annotation, he only has to show non-labeled items. This feature is already implemented. One of the advantages of this method is users can resume annotation when he/she use a different device from the one he/she has annotated. A disadvantage is it requires some implementation to the model, API and frontend.

c-w:
@Hironsan Thanks for your feedback and guidance. Would you like me to take a stab at implementing the localStorage approach? It would be a nice small incremental improvement.

clarus:
@Hironsan I would say that pagination is still useful to quickly skim over documents and get an overview of the text to annotate.

PR #199, opened 2019-05-15
Enhancement/Auto-generate label shortkeys and colors on corpus import

c-w:
In https://github.com/chakki-works/doccano/issues/5#issuecomment-433280629, @jamesmf brought up the suggestion to auto-generate shortkeys and colors for each imported label. This idea provides a great improvement to the user experience since the user can upload a partially annotated corpus and quickly start annotating efficiently afterwards thanks to all the shortkeys and colors already being assigned.

This pull request implements the automatic generation of label shortkeys and colors when a corpus is imported.

For each new label, we attempt to use the first character of the label text as the shortkey. If this is already set, we add modifier keys like ctrl, shift and ctrl+shift until we find a shortkey that doesn't yet exist. If all the shortkeys already exist, we move onto the next character in the label text and try again.

For each new label, we also auto-generate a background color and a foreground color as per the same algorithm that the frontend uses.

### DIFF
--- a/app/server/utils.py
+++ b/app/server/utils.py
@@ -4,6 +4,8 @@ import itertools
 import json
 import re
 from collections import defaultdict
+from math import floor
+from random import random
 
 from django.db import transaction
 from rest_framework.renderers import JSONRenderer
@@ -101,6 +103,11 @@ class BaseStorage(object):
                 serializer_label['prefix_key'] = shortkey[1]
                 existing_shortkeys.add(shortkey)
 
+            background_color = cls.generate_color()
+            text_color = cls.black_or_white(background_color)
+            serializer_label['background_color'] = background_color
+            serializer_label['text_color'] = text_color
+
             serializer_labels.append(serializer_label)
 
         return serializer_labels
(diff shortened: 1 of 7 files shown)


app/server/utils.py
```diff
@@ -74,16 +76,80 @@ def exclude_created_labels(self, labels, created):
...
+        suffix_keys = [key for key in label.lower() if key in model_suffix_keys]
+
+        for shortkey in itertools.product(suffix_keys, prefix_keys):
+            if shortkey not in existing_shortkeys:
+                return shortkey
+
+        return None
+
+    @classmethod
+    def generate_color(cls):
```
Hironsan: I think `generate_color` and `black_or_white` are implemented the same as one in JavaScript. But this may be better in a separate class or function like:

```python
import random


class Color(object):

    def __init__(self, rgb=None):
        self._rgb = rgb or random.sample(range(256), k=3)

    def contrast_color(self):
        """Generate black or white color.                                                                                                                              
                                                                                                                                                                       
        Ensure that foreground and background color combinations provide                                                                                               
        sufficient contrast when viewed by someone having color deficits or                                                                                            
        when viewed on a black and white screen.                                                                                                                       
                                                                                                                                                                       
        Algorithm from w3c:                                                                                                                                            
        * https://www.w3.org/TR/AERT/#color-contrast                                                                                                                   
        """
        r, g, b = self._rgb
        is_white = (((r * 299) + (g * 587) + (b * 114)) / 1000) < 128
        return Color([255] * 3) if is_white else Color([0] * 3)

    def to_hex(self):
        """Convert rgb to hex color."""
        return '#{:02x}{:02x}{:02x}'.format(*self._rgb)


if __name__ == '__main__':
    background_color = Color()
    text_color = background_color.contrast_color()
    print(background_color.to_hex())
    print(text_color.to_hex())
```

c-w: Done in c996d10.

PR #202, opened 2019-05-16
Enhancement/Move examples to unit tests

c-w:
There are some nice test examples of how to use the various classes in `utils.py` in the docstrings. However, doctests don't get executed by default by the Django test runner so we've had no programmatic guarantee that the examples keep up-to-date with the code. To fix this, this change moves the examples to normal unit tests. There is on functional change to the code.


c-w:
Working on merge conflicts.

PR #206, opened 2019-05-16
Bugfix/Encode username in Azure DATABASE_URL

c-w:
In the Azure deploy the database user is in the format `user@server` and it looks like furl doesn't deal well with that. URL-encoding the `@` to `%40` fixes the problem.

Resolves https://github.com/chakki-works/doccano/issues/203

### DIFF
--- a/azuredeploy.json
+++ b/azuredeploy.json
@@ -97,7 +97,7 @@
     "databaseSkuTier": "GeneralPurpose",
     "databaseSkuFamily": "Gen5",
     "databaseSkuName": "[concat('GP_', variables('databaseSkuFamily'), '_', parameters('databaseCores'))]",
-    "databaseConnectionString": "[concat('pgsql://', parameters('adminUserName'), '@', variables('databaseServerName'), ':', parameters('adminPassword'), '@', variables('databaseServerName'), '.postgres.database.azure.com:5432/', parameters('databaseName'))]",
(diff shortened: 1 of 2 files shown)


azuredeploy.json
```diff
@@ -97,7 +97,7 @@
     "databaseSkuTier": "GeneralPurpose",
     "databaseSkuFamily": "Gen5",
     "databaseSkuName": "[concat('GP_', variables('databaseSkuFamily'), '_', parameters('databaseCores'))]",
-    "databaseConnectionString": "[concat('pgsql://', parameters('adminUserName'), '@', variables('databaseServerName'), ':', parameters('adminPassword'), '@', variables('databaseServerName'), '.postgres.database.azure.com:5432/', parameters('databaseName'))]",
+    "databaseConnectionString": "[concat('pgsql://', parameters('adminUserName'), '%40', variables('databaseServerName'), ':', parameters('adminPassword'), '@', variables('databaseServerName'), '.postgres.database.azure.com:5432/', parameters('databaseName'))]",
```
cicorias: there's a template function `uricomponent(string)` that will do a URL encoded.

https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-template-functions-string#uricomponent

c-w: Given that we only have a single well-known character to encode and that we have `%40` in one of the tests, I'd wager to say that not using the function would be more understandable/readable here.

c-w: Done in 0811482.

PR #211, opened 2019-05-20
Enhancement/Alignment for import data format and export data format

erikamenezes:
Currently, the annotated file exported from doccano cannot be imported back into doccano for further validation or annotation because of a misalignment in the import and export formats described here (https://github.com/chakki-works/doccano/issues/113). 

This PR addresses this by adding a new export option that allows the users to export the annotated file with text labels that allows for this to be re-imported into doccano. The feature preview is shown below:
image

The original export functionality is maintained as it contains user details for each annotation that is used in inter-annotator analyses.

### DIFF
--- a/app/server/api.py
+++ b/app/server/api.py
@@ -209,13 +209,21 @@ class TextDownloadAPI(APIView):
         project = get_object_or_404(Project, pk=self.kwargs['project_id'])
         documents = project.documents.all()
         painter = self.select_painter(format)
-        data = painter.paint(documents)
+        
+        # json1 format prints text labels while json format prints annotations with label ids
+		# json1 format - "labels": [[0, 15, "PERSON"], ..]
+		# json format  "annotations": [{"label": 5, "start_offset": 0, "end_offset": 2, "user": 1},..]
+        if format == "json1":
+            labels = project.labels.all()
+            data = painter.paint_labels(documents, labels)
+        else:
+            data = painter.paint(documents)
         return Response(data)
 
     def select_painter(self, format):
         if format == 'csv':
             return CSVPainter()
-        elif format == 'json':
+        elif format == 'json' or format == "json1":
             return JSONPainter()
         else:
             raise ValidationError('format {} is invalid.'.format(format))
--- a/app/server/api.py
+++ b/app/server/api.py
@@ -211,8 +211,8 @@ class TextDownloadAPI(APIView):
         painter = self.select_painter(format)
         
         # json1 format prints text labels while json format prints annotations with label ids
-		# json1 format - "labels": [[0, 15, "PERSON"], ..]
-		# json format  "annotations": [{"label": 5, "start_offset": 0, "end_offset": 2, "user": 1},..]
+        # json1 format - "labels": [[0, 15, "PERSON"], ..]
+        # json format - "annotations": [{"label": 5, "start_offset": 0, "end_offset": 2, "user": 1},..]
         if format == "json1":
             labels = project.labels.all()
             data = painter.paint_labels(documents, labels)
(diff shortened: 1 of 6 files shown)


erikamenezes:
@Hironsan , @c-w :I'd really appreciate feedback on this PR. If I've missed something in the submitting process, please do let me know. 

PS: I'm new to the project and would love to contribute. (@c-w: Thanks for introducing me to this project!)

app/server/utils.py
```diff
@@ -391,7 +391,8 @@ def parse(self, file):
                 yield data
                 data = []
             try:
-                j = json.loads(line)
+                #j = json.loads(line)
+                j  = json.loads(line.decode('utf-8'))
```
c-w: Instead of calling `.decode` here, it may be better to wrap the entire file in a `TextIOWrapper` which will handle the encoding and ensure that we read the buffer correctly line by line. For an example, take a look at the [PlainTextParser implementation](https://github.com/chakki-works/doccano/blob/beeba225c7c916b3068337c244fafe487a356378/app/server/utils.py#L343).

app/server/api.py
```diff
@@ -209,13 +209,21 @@ def get(self, request, *args, **kwargs):
         project = get_object_or_404(Project, pk=self.kwargs['project_id'])
         documents = project.documents.all()
         painter = self.select_painter(format)
-        data = painter.paint(documents)
+        
+        # json1 format prints text labels while json format prints annotations with label ids
+        # json1 format - "labels": [[0, 15, "PERSON"], ..]
+        # json format - "annotations": [{"label": 5, "start_offset": 0, "end_offset": 2, "user": 1},..]
+        if format == "json1":
```
c-w: Instead of this conditional, should the `json1` format be a separate painter that extends or inherits from the JSONPainter? This will make the code more single-responsibility and will also make it easier in the future to evolve the two classes separately.

app/server/utils.py
```diff
@@ -391,7 +391,8 @@ def parse(self, file):
                 yield data
                 data = []
             try:
-                j = json.loads(line)
+                #j = json.loads(line)
```
c-w: Remove commented out code before merge.

app/server/utils.py
```diff
@@ -433,6 +433,24 @@ def paint(self, documents):
...
+        serializer = DocumentSerializer(documents, many=True)
+        data = []
+        for d in serializer.data:
+            labels = []
+            for a in d['annotations']:
+                label_obj = [x for x in serializer_labels.data if x['id'] == a['label']][0]
+                label_text = label_obj['text']
+                label_start = a['start_offset']
+                label_end = a['end_offset']
+                labels. append([label_start, label_end, label_text])
```
c-w: Formatting nit: no space before method name (`.append` not `. append`).

app/server/utils.py
```diff
@@ -433,6 +433,24 @@ def paint(self, documents):
             data.append(d)
         return data
 
+    def paint_labels(self, documents, labels):
+        serializer_labels = LabelSerializer(labels, many = True)
```
c-w: Formatting nit: no space between named argument and value (`many=True`, not `many = True`).

app/server/api.py
erikamenezes: I did consider that, but I thought it would be confusing to have two classes for JSON files. In this case my goal was for each function to cover a single responsibility. Also, given that there are other json types that people would like to export (e.g. Spacy) this class would handle all of them but have different functions for each format.  Thoughts?

erikamenezes:
Messed up some git history. See https://github.com/chakki-works/doccano/pull/217 for the updated PR.

PR #215, opened 2019-05-22
Enhancement/Document how to run tests

c-w:
Resolves https://github.com/chakki-works/doccano/issues/214

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,12 @@ Next we need to create a user who can login to the admin site. Run the following
 python manage.py create_admin --noinput --username "admin" --email "admin@example.com" --password "password"
 ```
 
+Developers can also validate that the project works as expected by running the tests:
+
+```bash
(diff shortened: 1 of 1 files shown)


erikamenezes:
Suggestion: Add collectstatic as well?

Also, even though the django tests pass, the CI sometimes breaks due to some style guidelines. So I've found running the tools/ci.sh to be better.

c-w:
The need to run collectstatic before the tests will be removed as soon as https://github.com/chakki-works/doccano/pull/207 is merged.

I'm hesitant to recommend `tools/ci.sh` as the default test runner since it won't work on Windows. Ideally we'd re-write the bash scripts in Python so that they can be run cross-platform. I can work on that.

Hironsan:
It is better to explain only how to run the basic test. It's executable for any developers.

PR #216, opened 2019-05-22
Feature/Add option to randomize document order per user

c-w:
In https://github.com/chakki-works/doccano/issues/201, @rbagd mentioned that having an option to randomize document order per user is a great way to reduce annotation fatigue. This pull request adds a new field to the project model which enables this optional feature.

Screenshot showing document order randomization method on project creation page


c-w:
@rbagd I'd appreciate if you could try out this branch and let me know if this fits your requirements and/or what improvements we could make to this feature. Thanks in advance!

rbagd:
I just tested this and it looks pretty great! Simple, yet effective.

What I like in this simple soluton is that although it can look somewhat random, it is still pretty easy for us to check whether annotation fatigue has a statistical effect after the tagging exercise.

PR #217, opened 2019-05-23
Enhancement/Alignment for import data format and export data format 

erikamenezes:
Currently, the annotated file exported from doccano cannot be imported back into doccano for further validation or annotation because of a misalignment in the import and export formats described here (https://github.com/chakki-works/doccano/issues/113). 

This PR addresses this by adding a new export option that allows the users to export the annotated file with text labels that allows for this to be re-imported into doccano. The feature preview is shown below:
image

The original export functionality is maintained as it contains user details for each annotation that is used in inter-annotator analyses.


erikamenezes:
I messed up the git history in the previous PR (https://github.com/chakki-works/doccano/pull/211). Please see this instead. cc: @Hironsan , @c-w

PR #232, opened 2019-06-06
Import and export using SpaCy NER format

AI-ITVT:
Import and export sequence labeling annotations in SpaCy NER training data format.
https://spacy.io/usage/spacy-101#annotations-ner


Hironsan:
Need to pass tests.

Issue #236, opened 2019-06-08
Bug: python app/manage.py collectstatic takes huge time.


icoxfog417:
### System information

Any platform

### Describe the problem

`python app/manage.py collectstatic` takes huge time.
If may be caused by [Adopt nuxt folder structure](https://github.com/chakki-works/doccano/commit/9206b499c23f6d88b9d6963cc8ae15559ec1783b#diff-22ec2e0f0766e1f9083041c2fa1e3c47).

* Before: `node_modules` folder is located at `/src/app/server` (outside of `static` folder)
* Ater: `node_modules` folder is located **inside** `/src/app/server/static`.

It causes the following problems.

* Increase size of the docker container.
* Timeout error on Heroku deploy.

### Source code / logs

`python app/manage.py collectstatic` log.

Before (When #101)

```
$ python app/manage.py collectstatic --noinput
       217 static files copied to '/tmp/build_9b389474fc695765e959a7fec753ed34/app/staticfiles', 928 post-processed.
```

After

```
$ python app/manage.py collectstatic --noinput
       25097 static files copied to '/tmp/build_a28a2c009c7b6a54a25ae7f6e7059e6e/app/staticfiles', 118207 post-processed.
```

static files are from 217 =>  25097.

PR #237, opened 2019-06-08
[WIP] Bugfix/Fix collectstatic target (#236)

icoxfog417:
Fix collectstatic target (#236)

[Deploy](https://heroku.com/deploy?template=https://github.com/chakki-works/doccano/tree/issue_236)

**Heroku Deploy is testing now**

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.pythonPath": "venv/bin/python3.7"
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ RUN cd /doccano/app/server/static \
  && rm -rf components pages node_modules .*rc package*.json webpack.config.js
 
 RUN cd /doccano \
- && python app/manage.py collectstatic --noinput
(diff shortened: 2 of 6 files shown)


icoxfog417:
We have to set [ALLOWED_HOSTS](https://docs.djangoproject.com/es/2.1/ref/settings/#allowed-hosts) to get static files on the production environment (`DEBUG=False`).

When `DEBUG=True`, ALLOWED_HOSTS is automatically set to localhost, etc. But when `DEBUG=False`, we have to specify it explicitly.

c-w:
@icoxfog417 Would it be possible/desirable to change the Heroku deployment to use Docker to prevent maintaining two deployment modes? What would be the drawbacks of this approach?

icoxfog417:
@c-w You totally have a point. Heroku supports docker deploy but it is not supported at Heroku Button (=To support one-click deployment, we have to use default deploy). It is only and most important problem.

icoxfog417:
Apart from that, `ALLOWED_HOSTS` problem should be solved. We encounter "Server Error 500" when `DEBUG=False` even if localhost (local development environment).

PR #250, opened 2019-06-18
Authentification

guillim:
1. This PR is mainly for setting up the self-authentification from the user side. Basically, there is a ENV variable called "**ALLOW_SIGNUP**" in the **docker-compose.yml** file. It's a boolean:
- True : users can create their account on the plateform  
- False : only the admin can create accounts for the users

2. The second feature is for resetting the password


guillim:
@Hironsan Sorry to bother you, but I don't understand how to fix [this codacy issue](https://app.codacy.com/app/Hironsan/doccano/pullRequest?prid=3735569)  
 `Unexpected keyword argument 'backend' in function call` from the file _app/authentification/utils.py_ 

If I remove the **backend** argument (as suggested by codacy), then I have a python error, asking for a **backend** specification...
```
You have multiple authentication backends configured and therefore must provide the `backend` argument or set the `backend` attribute on the user.
```

Hironsan:
@guillim I think the problem is testing coverage.

According to [Travis CI's build history](https://travis-ci.com/chakki-works/doccano/builds/116413274),  the coverage score is 89%. By https://github.com/chakki-works/doccano/pull/205, if the coverage drops below 90%, the CI process fails.

guillim:
Thanks for your answer, I actually didn't write any test yet, and so I will do !

guillim:
That's it:
- Tests are set up properly 
- Coverage is **92%**  (which is higher than the 90% threshold)
- Small typos fixed

app/app/settings.py
```diff
@@ -85,7 +88,7 @@
 TEMPLATES = [
     {
         'BACKEND': 'django.template.backends.django.DjangoTemplates',
-        'DIRS': [path.join(BASE_DIR, 'server/templates')],
+        'DIRS': [path.join(BASE_DIR, 'server/templates'),path.join(BASE_DIR, 'authentification/templates')],
```
Hironsan: Need a space after a comma.

docker-compose.yml
```diff
@@ -12,6 +12,7 @@ services:
       ADMIN_PASSWORD: "password"
       ADMIN_EMAIL: "admin@example.com"
       DATABASE_URL: "postgres://doccano:doccano@postgres:5432/doccano?sslmode=disable"
+      ALLOW_SIGNUP: "True"
```
Hironsan: As a default value, I think "False" is better than "True" because it prevents unexpected signup.

app/authentification/forms.py
```diff
@@ -0,0 +1,11 @@
+from django import forms
+from django.contrib.auth.forms import UserCreationForm
+from django.contrib.auth.models import User
```
Hironsan: `from django.contrib.auth import get_user_model` is a better way to refer to User. The reason is:
* https://docs.djangoproject.com/en/2.2/topics/auth/customizing/#referencing-the-user-model

app/authentification/utils.py
```diff
@@ -0,0 +1,23 @@
+from django.shortcuts import render
+from django.shortcuts import redirect
+from django.contrib.auth import login
+from django.utils.encoding import force_text
+from django.utils.http import urlsafe_base64_decode
+from .tokens import account_activation_token
+from django.contrib.auth.models import User
```
Hironsan: `from django.contrib.auth import get_user_model` is a better way to refer to User. The reason is:
* https://docs.djangoproject.com/en/2.2/topics/auth/customizing/#referencing-the-user-model

Hironsan:
@guillim Thank you for your great work. I reviewed some code.

app/app/settings.py
guillim: PEP8 yeah

docker-compose.yml
guillim: Fine for me

app/authentification/forms.py
guillim: True, definitely a better way.

app/authentification/utils.py
guillim: same thing as app/authentification/forms.py

guillim:
> @guillim Thank you for your great work. I reviewed some code.

Just applied the required changes, and merged with your latest work! @Hironsan

Issue #253, opened 2019-06-21
When the number of labels is more than 15, colors are difficult to distinguish


Juntai:
The UI of doccano is very beautiful! 
I have a project of sequence labeling with many labels, about 20 labels. doccano use COLOR to distinguish labels, but it is difficult to distinguish labels when there are too many labels.
I know I can use shortcut to label them, but when I review the labeled sequences to correct them, it is very difficult to distinguish labels.
![image](https://user-images.githubusercontent.com/4479020/59905776-a8599280-9439-11e9-9f4c-06653f8570bd.png)

There may be one way to solve that:
Add label names around the sequences text, left or above it.

clarus:
Just having the same problem! I was thinking about adding a tooltip under the annotations:
![image](https://user-images.githubusercontent.com/316665/59909267-906d1900-940f-11e9-8953-925a76cf11a9.png)

Juntai:
Thank you , clarus. I think this style may be better: https://prodi.gy/demo?view_id=pos_manual

clarus:
Ah OK, I did not want to make too much change to the code and the presentation.

PR #254, opened 2019-06-21
Add a tooltip under each annotation to name them

clarus:
This would help to distinguish annotations when there are too many labels to distinguish the colours. Related to the issue https://github.com/chakki-works/doccano/issues/253
image

### DIFF
--- a/app/server/static/components/annotator.vue
+++ b/app/server/static/components/annotator.vue
@@ -2,7 +2,8 @@
   div(@click="setSelectedRange")
     span.text-sequence(
       v-for="r in chunksWithLabel"
-      v-bind:class="{ tag: id2label[r.label].text_color }"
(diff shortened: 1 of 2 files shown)


Juntai:
Thank you , clarus.  I submit this issue [#253](https://github.com/chakki-works/doccano/issues/253). 
Is the tooltip you added a mouse hovering?
 If that is true, I think this style may be better: https://prodi.gy/demo?view_id=pos_manual

clarus:
Yes, this is mouse hovering.

Juntai:
Thank you. Would you consider this solution?
TimLine截图20190621170641

clarus:
Not for now.

Hironsan:
Thanks!

Issue #123, opened 2019-03-25
New project structure


Hironsan:
This issue is the first step to the *Introduce JavaScript testing framework* task I wrote in the [Feature Ideas](https://github.com/chakki-works/doccano/projects/2#card-16405365).

I think the current project structure is somewhat hard to understand for both Django and Vue.js developers and difficult to decide where to place the frontend testing code. To make it easier to understand project structure and to place the testing code, the project structure should be changed to something more clear like follows:

```
doccano/
  backend/
    app/
      __init__.py
      settings.py
      urls.py
      wsgi.py
    api/         # doccano core api; Project, Documents, ...
      tests/
      admin.py
      exceptions.py
      filters.py
      models.py
      permissions.py
      serializers.py
      urls.py
      utils.py
      views.py
    router/ 
      templates/
      static/
      tests/
      apps.py
      urls.py
      views.py
    management/
    __init__.py
    manage.py
  frontend/  # Client Side Vue/Javascript, build could be written to static/ folder
    src/
      assets/
      components/
    tests/
      units/
    package.json
    package-lock.json
    webpack.config.js
  docs/
  scripts/
  Pipfile
  Pipfile.lock
  ...
```

This structure is based on [cookiecutter-django-vue](https://github.com/vchaptsev/cookiecutter-django-vue).

frco9:
@Hironsan, could you clarify the status of this refactorization? Shall we use the code from the `frontend` directory or update components is the `app/server/statics` directory? 
I'd like to do some improvements and a PR but this point is not very clear for me? !

PR #256, opened 2019-06-25
Enhancement/Separate a Django application into API app and router app

Hironsan:
I think the current backend structure is somewhat hard to understand because it has two roles: APIs and router. To make it easier to understand project structure, the project structure should be changed  like follows:

```
doccano/
  app/
    app/
      __init__.py
      settings.py
      urls.py
      wsgi.py
    api/
      migrations/
      tests/
      admin.py
      apps.py
      exceptions.py
      filters.py
      models.py
      permissions.py
      serializers.py
      urls.py
      utils.py
      views.py
    server/ 
      templates/
      templatetags/
      static/
      tests/
      apps.py
      forms.py
      social_auth.py
      urls.py
      views.py
```


icoxfog417:
I think the name "app" and "server" is ambiguous and `app` folder does not have an explicit role.
So I propose to flatten the `app` folder and split features to backend/frontend (or api/site).

[doccano]
├── [doccano]
│          ├ __init__.py
│          ├  urls.py
│          ├  wsgi.py
│          ├  settings.py
│          ├  [backend (api)]
│          └  [frontend (site)]
├── [docs]
└── [tools]

Hironsan:
In this issue, I focus on separating a Django application. So I don't want to do many things at the same time. I have already designed project structure as follows:

* https://github.com/chakki-works/doccano/issues/123

icoxfog417:
If this PR is a step toward #123, we should re-open #123 (or create another issue) because it is not done yet and the other contributors can't know structure change is still in progress.

PR #260, opened 2019-06-26
Bugfix/wrong lable

Alan-buaa:
if use doc.index() to find text position, when this entity is appeared before this words, we find wrong position.


Hironsan:
@Alan-buaa Please tell me some specific case.

PR #264, opened 2019-06-27
Feature/spacy import export

AI-ITVT:
Support data import+export in SpaCy training data format.

- now added unit tests to keep overall CI test coverage above 90%.
  For someone new to this project, an error message regarding this threshold would be useful. It took me quite a while to figure why CI failed.

### DIFF
--- a/app/server/utils.py
+++ b/app/server/utils.py
@@ -363,7 +363,7 @@ class SpacyParser(FileParser):
     def parse(self, file):
         file = io.TextIOWrapper(file, encoding='utf-8')
         # needs loading of full document
-        items = json.loads(file.read()) 
+        items = json.loads(file.read())
         data = []
         for i, line in enumerate(items, start=1):
             if len(data) >= IMPORT_BATCH_SIZE:
--- a/app/server/utils.py
+++ b/app/server/utils.py
@@ -358,6 +358,25 @@ class JSONParser(FileParser):
         if data:
             yield data
 
+class SpacyParser(FileParser):
+
+    def parse(self, file):
+        file = io.TextIOWrapper(file, encoding='utf-8')
+        items = json.loads(file.read())  # needs loading of full document 
+        data = []
+        for i, line in enumerate(items, start=1):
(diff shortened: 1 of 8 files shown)


app/server/utils.py
```diff
@@ -405,6 +425,26 @@ def paint(self, documents):
...
+            entities = []
+            for a in annotations:
+                label_id = a["label"]
+                label_text = Label.objects.get(id=label_id).text
+                entities.append([a["start_offset"], a["end_offset"], label_text])
+            res.append( tuple([text, {"entities" : entities}]) )
+        return [res]
+
+
+
```
c-w: Nit: only two newlines between classes.

app/server/utils.py
```diff
@@ -405,6 +425,26 @@ def paint(self, documents):
...
+    def paint(self, documents):
+        data = super().paint(documents)
+        res = []
+        for d in data:
+            text = d['text']
+            annotations = d.pop('annotations')
+            entities = []
+            for a in annotations:
+                label_id = a["label"]
+                label_text = Label.objects.get(id=label_id).text
```
c-w: This will do a lot of calls to the database: number of data lines times number of annotations per data line.

Ideally we'd do a single trip to the database that fetches all the relevant label texts. Something along these lines perhaps?

```py
label_ids = {a['label'] for d in data for a in d['annotations']}
label_texts = {l['id']: l['text'] for l in Label.objects.filter(id__in=label_ids).values('text', 'id')}

for d in data:
  # ...
  for a in annotations:
    label_text = label_texts[a['label']]
```

(Sample code is just suggested as a starting point and wasn't tested or run.)

app/server/utils.py
```diff
@@ -405,6 +425,26 @@ def paint(self, documents):
...
+                entities.append([a["start_offset"], a["end_offset"], label_text])
+            res.append( tuple([text, {"entities" : entities}]) )
```
c-w: Nit: call to `tuple` is redundant and can be simplified to `res.append((text, {"entities" : entities}))`.

PR #269, opened 2019-07-01
Enhancement/Add viewer pane for source document

c-w:
In many situations, users will use Doccano to annotate text that was extracted from documents such as PDF, Word or via OCR. Often times, the plain text extraction process will reduce the amount of information available to the annotator, e.g. loss of formatting, loss of page breaks, etc. Being able to view the source document during the annotation process therefore often is helpful to the annotator.

As such, this pull request adds a preview pane for PDF and Word documents to the annotation pane. The preview pane is toggled on the document level by adding the `documentSourceUrl` key in the metadata ([sample corpus to test the document viewer pane](https://github.com/chakki-works/doccano/files/3347079/example.jsonl.txt)).

Animation showing source document viewer pane

### DIFF
--- a/app/server/static/components/annotation.pug
+++ b/app/server/static/components/annotation.pug
@@ -133,7 +133,12 @@ div.columns(v-cloak="")
           span.icon
             i.fas.fa-box
 
-    block annotation-area
+    div.columns
+      div.column
+        block annotation-area
+
+      div.column(v-if="documentMetadata != null && documentMetadata.documentSourceUrl != null")
+        preview(v-bind:url="documentMetadata.documentSourceUrl")
 
     div.level.mt30
       a.button.level-left(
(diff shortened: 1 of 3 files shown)


guillim:
@psorianom this could be very interesting for your use case

c-w:
@guillim @psorianom Could you tell me a bit more about your use-cases so that we can make sure to consider them in this implementation? Thanks in advance!

guillim:
Hello @c-w,
We are working on some __legal case__ annotation. Such texts have very specific indentation & paragraph styling: it matters to understand the meaning of a text. 

In my opinion, annotators would really benefit from such a feature. However, it could even be better if we could have a "markdown like" way to record texts, so that displaying line breaks in a text might be possible (instead of a __text__ + __text preview__)

c-w:
Docker build failure is due to https://github.com/chakki-works/doccano/pull/290, unrelated to the changes in this pull request.

Paooooo:
Hello, 

I am very interested in this preview panel feature for pdf files.
I uploaded the sample jsonl file with `documentSourceUrl` key in the metadata as it is said it will toggle the preview pane, but in my case it is not.

Am I missing some extra steps to toggle this preview panel ?

c-w:
@Paooooo I believe this feature is only available in the v0 UI, not in the newer v1 UI.

Paooooo:
@c-w Ok, thanks for the precision.

PR #273, opened 2019-07-04
Add highlight of the search results

clarus:
Add a yellow highlight to show where the search results are in the text.

image

One disadvantage of this implementation is that it adds `span`s to the text, but I did not find how to do otherwise.

### DIFF
--- a/app/server/static/components/annotator.vue
+++ b/app/server/static/components/annotator.vue
@@ -8,7 +8,12 @@
         color: id2label[r.label].text_color, \
         backgroundColor: id2label[r.label].background_color \
       }"
-    ) {{ textPart(r) }}
+    )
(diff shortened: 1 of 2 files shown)


Hironsan:
Thank you for your PR, @clarus.

I think [mark.js](https://markjs.io/) is another choice to implement this feature. By using this, the code may be simpler. Below is an example:

* https://jsfiddle.net/julmot/vpav6tL1/

clarus:
Thanks for the suggestion. Actually I did look into [mark.js](https://markjs.io/) but though it would not be appropriate as jQuery plugins may not interact well with Vue.js VDOM and philosophy.

clarus:
Thanks!

PR #276, opened 2019-07-05
Enhancement/Remove demo components

c-w:
As discussed in https://github.com/chakki-works/doccano/issues/274, we currently have some code duplication between the demo labeling canvas and the production labeling canvas. This means that new frontend features may have to be implemented twice which slows down the process and may confuse developers.

This pull request refactors the demo pages to use the production labeling canvas. This is achieved by using [axios-mock-adapter](https://github.com/ctimmerm/axios-mock-adapter) to intercept the HTTP requests of the production labeling canvas and replacing them with an in-memory implementation. In the future, this same approach can also be leveraged for faking HTTP requests for unit testing the frontend.


Hironsan:
Wonderful!

PR #285, opened 2019-07-11
added environment.yml and updated readme

hkaraoguz:


environment.yml
```diff
@@ -0,0 +1,89 @@
...
+    - tornado==5.0.2
+    - unicodecsv==0.14.1
+    - unittest-xml-reporting==2.5.1
+    - urllib3==1.24.3
+    - vcrpy==2.0.1
+    - vcrpy-unittest==0.1.7
+    - whitenoise==4.1.2
+    - wrapt==1.11.2
+    - yarl==1.3.0
+prefix: /home/hakan/anaconda3/envs/doccano
```
c-w: This looks like a line that shouldn't be checked in since it'll be different for each developer. Is there any way to make this generic?

hkaraoguz: Hi. As I read from here https://stackoverflow.com/questions/41274007/anaconda-export-environment-file conda does not care about that line but puts it there anyway while exporting. I removed the line and built the environment without any issues.

PR #286, opened 2019-07-11
Enhancement/elasticsearch

margaretmeehan:
this simply adds Elasticsearch option to search documents during annotation. If the .env file has an Elasticsearch host specified, then the Document List will populate using Elasticsearch results. 

The document is simply indexed on the text field, as that is what we want to search on. That's currently how the filtering works for Documentlist.


app/api/views.py
```diff
@@ -140,7 +140,16 @@ class DocumentList(generics.ListCreateAPIView):
     def get_queryset(self):
         project = get_object_or_404(Project, pk=self.kwargs['project_id'])
 
-        queryset = project.documents
+        if (settings.ELASTIC_SEARCH_HOST):  # Elasticsearch
+            from .index import DocumentIndex
```
margaretmeehan: I know this import shouldn't go here ... but anyone have recommendations on conditional imports? I only want this Index to be loaded in if we are using Elasticsearch.

Issue #268, opened 2019-06-29
Sequence to sequence: cannot download labels


raibosome:
### System information
- macOS 10.14
- Python 3.6

### Describe the problem
Ran the Docker image as a container. Created a sequence to sequence project and imported a one-column .txt file. Managed to annotate the data and the annotations persist but I notice that:

1.  Progress bar shows 0/0 despite having imported data.
2. No statistics available.
3. Cannot export data.

### Source code / logs

[Here](https://github.com/chakki-works/doccano/files/3341857/testing.txt) is the data that I imported. Below is a screenshot that shows the progress bar is at 0/0.

<img width="1674" alt="image" src="https://user-images.githubusercontent.com/11023859/60383923-923b7a00-9aaa-11e9-8885-d84a83ccbf18.png">

Logs for viewing statistics.

```
Internal Server Error: /v1/projects/4/statistics
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 126, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 124, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/views/generic/base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 483, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 443, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 480, in dispatch
    response = handler(request, *args, **kwargs)
  File "/doccano/app/api/views.py", line 70, in get
    label_count, user_count = self.label_per_data(p)
  File "/doccano/app/api/views.py", line 94, in label_per_data
    for d in annotations.values('label__text', 'user__username').annotate(Count('label'), Count('user')):
[...]

PR #287, opened 2019-07-12
Bugfix/seq2seq label download

fnocera:
Addresses issue #268 

- Fixes progress bar so that it shows number of labelled documents for seq2seq. `label_per_data` now implemented through an `AnnotationManager`. 
- Added field that was missing/causing an error in the output to allow for seq2seq export.

Seq2Seq-bugfix

### DIFF
--- a/app/api/models.py
+++ b/app/api/models.py
@@ -7,6 +7,8 @@ from django.contrib.staticfiles.storage import staticfiles_storage
 from django.core.exceptions import ValidationError
 from polymorphic.models import PolymorphicModel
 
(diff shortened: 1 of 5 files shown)


guillim:
Sounds like a good fix for an important bug

app/api/models.py
```diff
@@ -225,6 +229,9 @@ class Meta:
 
 
 class Seq2seqAnnotation(Annotation):
+    # Override AnnotationManager for custom functionality
```
c-w: I'd remove this comment since it's implied by overriding of the object.

PR #291, opened 2019-07-16
Added color dropdown to help choose contrasting colors for the labels.

fnocera:
Addressing chakki-works#175 by adding a dropdown of contrasting default colors. Randomizer button is still available and text box color picker is within the dropdown. 

Color gif 2

### DIFF
--- a/app/server/static/components/label.vue
+++ b/app/server/static/components/label.vue
@@ -160,7 +160,7 @@
                           v-bind:trigger-style="{ width: '36px', height: '36px' }"
                         )
                   div.control
-                    a.button.random-color-button-edit(
+                    a.button.random-color-button(
                       v-on:click="setEditColor"
                     )
                       span.icon.is-small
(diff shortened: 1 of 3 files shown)


app/server/static/components/label.vue
```diff
@@ -176,13 +177,33 @@
...
+<style scoped>
+.random-color-button {
+  height: 36px;
+  width: 36px;
+  background-color: transparent;
+  color: #404040;
+  border: none;
+}
+
+.random-color-button-edit {
```
c-w: This has the same CSS rules as the `.random-color-button` selector. Can the two selectors be merged? Either reference the same `.random-color-button` class in the template for both elements, or merge the CSS definitions:

```css
.random-color-button, .random-color-button-edit {
 ...
}
```

fnocera: Thanks for spotting this - now both uses reference the same `.random-color-button` class.

c-w:
@fnocera Please rebase from master since the Dockerfile changed.

PR #298, opened 2019-07-19
Use library conllu

shizacat:
Hello,
It may be better to be closer to the standard. And use the read library.


requirements.txt
```diff
@@ -36,3 +36,4 @@ unittest-xml-reporting==2.5.1
 vcrpy==2.0.1
 vcrpy-unittest==0.1.7
 whitenoise[brotli]==4.1.2
+conllu
```
Hironsan: which version?

app/api/tests/test_utils.py
```diff
@@ -143,13 +143,16 @@ def test_make_annotations(self):
...
+        f = io.BytesIO()
 
-        entities = get_entities(tags)
-        actual = CoNLLParser.calc_char_offset(words, tags)
+        s = [
+            ("EU", "ORG"), ("rejects", "_"), ("German", "MISC"), ("call", "_")
+        ]
+        for w, t in s:
+            f.write("{}\t{}\n".format(w, t).encode())
+        f.seek(0)
```
c-w: In order to avoid these one-letter variable names, I wonder if it would be simpler to just hard-code the test string like this?

```py
f = io.BytesIO(
  b"EU\tORG\n"
  b"rejects\t_\n"
  b"German\tMISC\n"
  b"call\t_\n"
)
```

requirements.txt
c-w: Should this be pinned to `conllu==1.3.1`?

app/api/utils.py
```diff
@@ -242,45 +243,51 @@ class CoNLLParser(FileParser):
...
-            line = line.strip()
-            if line:
-                try:
-                    word, tag = line.split('\t')
-                except ValueError:
-                    raise FileParseException(line_num=i, line=line)
+
+        # Add check exception
+
+        field_parsers = {
```
c-w: Nit: can be inlined.

app/api/utils.py
```diff
@@ -242,45 +243,51 @@ class CoNLLParser(FileParser):
...
+            "ne": lambda line, i: conllu.parser.parse_nullable_value(line[i]),
+        }
+
+        try:
+            sentences = conllu.parse(
```
c-w: This should probably use `parse_incr` in order to avoid loading the entire file into memory.

app/api/utils.py
```diff
@@ -242,45 +243,51 @@ class CoNLLParser(FileParser):
...
+                fields=("form", "ne"),
+                field_parsers=field_parsers
+            )
+        except conllu.parser.ParseException as e:
+            raise FileParseException(line_num=-1, line=str(e))
+
+        for sentence in sentences:
+            if not sentence:
+                continue
+            # if len(data) >= settings.IMPORT_BATCH_SIZE:
```
c-w: Why did this need to be removed? We should keep this functionality to ensure that we're importing in the correct batch size.

app/api/utils.py
```diff
@@ -242,45 +243,51 @@ class CoNLLParser(FileParser):
...
+            #     yield data
+            #     data = []
+            words, labels = [], []
+            for item in sentence:
+                word = item.get("form")
+                tag = item.get("ne", None)
```
c-w: Nit: `.get('ne', None)` should be equivalent to `.get('ne')` for any normal dictionary.

requirements.txt
shizacat: Yes, is 1.3.1

app/api/tests/test_utils.py
shizacat: Yes, it is really prettier

app/api/utils.py
shizacat: This will lengthen the design.

app/api/utils.py
shizacat: Not sure if there is a need

app/api/utils.py
```diff
@@ -242,45 +243,51 @@ class CoNLLParser(FileParser):
...
+                    if tag is not None:
+                        char_left = sum(map(len, words)) + len(words)
+                        char_right = char_left + len(word)
+                        span = [char_left, char_right, tag]
+                        labels.append(span)
+
+                    words.append(word)
+
+                # Create JSONL
+                j = {'text': ' '.join(words), 'labels': labels}
```
c-w: Nit: I'd inline this variable to avoid short/non-descriptive variable-names like `j`.

Issue #81, opened 2019-01-29
Feature Request: Question Answering dataset


jamesmf:
### Describe the problem
Given the relative scarcity of tools for building a question-answering dataset, it would be great if doccano could serve that purpose. At its core, I believe that would mean each document is a passage, then there would be forms below (like in a Translation project) that allow you to enter in questions. Then you'd need the ability to annotate stretches of text as the answer (similar to a Sequence Labeling project). The main functionality that would be tricky would likely be tying the labels for the answer annotation back to the input questions.

A naive first pass could be to replicate a combination of the SequenceLabeling+MachineTranslation UIs together. The Labels for each answer could be simply `answer-to-Q1`, `answer-to-Q2` and each document could simply have a max number of questions associated.

Does this sound doable?

icoxfog417:
Thank you for the feature request! At first, please let me arrange the type of QA related dataset. Because there are many variations.

**Passage type**

1. Single-passage type: Only one passage is shown.
2. Multi-passage type: Multiple passages are shown.
   * QAngaroo,  HotpotQA, MultiRC etc
3. Multimedia passage type: Passage includes not only text, but also image etc.
   * VQA, FigureQA etc

**Answer type**

1. Extractive answer type: The answer is a span in a passage.
   * SQuAD, Maluuba News QA etc
2. Selective answer type: Select the answer from candidates.
   * AI2 Reasoning Challenge Dataset, FigureQA etc
3. Generative answer type: The answer is generated by a human.
   * MSMARCO V2, MilkQA etc

Fow now, you can use doccano in following situation.

* By Sequence to Sequence annotation feature
  * Passage type is 1 & Answer type is 3. (Show passage and question as annotation target).
* By Text Classificatioin feature
  * Passage type is 1 & Answer type is 2. (Show passage/question/choices as annotation target).

We want to decide from user feedback that what kind of QA dataset is appropriate to support.
So please send us your comments!

guillim:
> My use case is `1&1` -- SQuAD-like.
> 
> The option for Passage-type 2 seems cool and achievable. The option for Passage-type 3 would be very cool, but seems like it requires turning doccano into a multimodal annotation tool (a great goal!) which feels farther off.
> 
> I could see wanting to eventually use any of the Answer-types, but I agree that it is definitely possible to set up Answer-types 2 & 3 with current functionality.

Any update about your question @jamesmf ? I am looking for the same thing: a way to reproduce SQuAD in another language...

[...]

PR #304, opened 2019-07-23
New project type: Question and answer

guillim:
**Summary:**
This pull request has a simple objective: creating a new annotation mode, for the task "question and answer" 

Ref: #81 

**What changed**

- Backend: 2 models were created (_QandAProject_  & _QandAAnnotation_ ) and all the following related (serializer)
The model _Document_ was slightly modified, adding the text field called _extra_text_ - allows some tasks to store additional information, like a question for instance.
Migrations were created following the model changes as well
Questionable: the necessity of having QandAStorage. At the moment, I guess I could use _Seq2seqStorage_ instead.  

- Frontend: a new front was added using the component _download_qanda.vue_. Visual examples for uploading / downloading data were created as well.

**Other Info**
This is a Draft pull request at the moment. No tests were created yet, and it isn't merged with the latest master branch due to a problem with _pyodbc_ on my side


guillim:
@Hironsan If you have time to review this, it would be very nice. From the issue #81 it sounded like many people were waiting for this feature. Let me know if you see improvement I could do!

app/server/static/package-lock.json
```diff
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "7.0.0"
+        "@babel/highlight": "^7.0.0"
```
c-w: Given that `package.json` didn't change as part of this pull request, can we revert the changes to `package-lock.json` since they don't pertain to the changes in this pull request?

app/server/static/components/question_and_answer.vue
```diff
@@ -0,0 +1,78 @@
...
+
+block annotation-area
+  div.card
+    header.card-header
+      div.card-header-title.has-background-royalblue
+        div.field.is-grouped.is-grouped-multiline
+          div.control
+            div.tags.has-addons.white(
+              v-if="docs[pageNumber] && docs[pageNumber].extra_text"
+              ) {{ docs[pageNumber].extra_text || 'missing quesiton' }}
```
c-w: Nit: typo `missing quesiton` should be `missing question`.

c-w: Nit: this should be indented like so:

```pug
            div.tags.has-addons.white(
              v-if="docs[pageNumber] && docs[pageNumber].extra_text"
            ) {{ docs[pageNumber].extra_text || 'missing quesiton' }}
```

Basically there are two extra spaces before the `)`.

app/server/static/components/question_and_answer.vue
```diff
@@ -0,0 +1,78 @@
...
+          v-bind:entity-positions="annotations[pageNumber]"
+          v-bind:text="docs[pageNumber].text"
+          v-on:remove-answer="removeLabel"
+          v-on:add-answer="addAnswer"
+          ref="annotator"
+        )
+    div.card-content.has-text-centered
+      button.button.is-link.is-medium(
+        v-bind:disabled="this.hasAnswer"
+        v-on:click="annotate()") Submit Selection
```
c-w: Nit: to be consistent with other formatting, this should look like so:

```pug
      button.button.is-link.is-medium(
        v-bind:disabled="this.hasAnswer"
        v-on:click="annotate()"
      ) Submit Selection

```

app/server/static/components/annotator_qanda.vue
```diff
@@ -0,0 +1,179 @@
...
+    startOffset: 0,
+    endOffset: 0,
+  }),
+
+  computed: {
+    sortedEntityPositions() {
+      /* eslint-disable vue/no-side-effects-in-computed-properties */
+      this.entityPositions = this.entityPositions.sort((a, b) => a.start_offset - b.start_offset);
+      return this.entityPositions;
+      /* eslint-enable vue/no-side-effects-in-computed-properties */
```
c-w: Why is it necessary to have the side-effect here? Why not simply return `this.entityPositions.sort((a, b) => a.start_offset - b.start_offset);`?

app/server/static/components/annotator_qanda.vue
```diff
@@ -0,0 +1,179 @@
...
+    },
+
+    chunks() {
+      const res = [];
+      let left = 0;
+      let e;
+      for (let i = 0; i < this.sortedEntityPositions.length; i++) {
```
c-w: Can this be simplified with a for-of loop?

```js
for (let e of this.sortedEntityPositions) {
  // ...
}
```

app/server/static/components/annotator_qanda.vue
```diff
@@ -0,0 +1,179 @@
...
+      /* eslint-disable vue/no-side-effects-in-computed-properties */
+      this.entityPositions = this.entityPositions.sort((a, b) => a.start_offset - b.start_offset);
+      return this.entityPositions;
+      /* eslint-enable vue/no-side-effects-in-computed-properties */
+    },
+
+    chunks() {
+      const res = [];
+      let left = 0;
+      let e;
```
c-w: Would `entity` be a better name for this variable?

app/server/static/components/annotator_qanda.vue
```diff
@@ -0,0 +1,179 @@
...
+      for (let i = 0; i < this.sortedEntityPositions.length; i++) {
+        e = this.sortedEntityPositions[i];
+        e.label = 1;
+        e.end_offset = e.start_offset + e.response.length;
+
+        const l = this.makeEmptyChunk(left, e.start_offset);
```
c-w: Should this variable be called `label` or some-such?

app/server/static/components/annotator_qanda.vue
```diff
@@ -0,0 +1,179 @@
...
+        const selectedTextRange = document.selection.createRange();
+        const preSelectionTextRange = document.body.createTextRange();
+        preSelectionTextRange.moveToElementText(this.$el);
+        preSelectionTextRange.setEndPoint('EndToStart', selectedTextRange);
+        start = [...preSelectionTextRange.text].length;
+        end = start + [...selectedTextRange.text].length;
+      }
+      this.startOffset = start;
+      this.endOffset = end;
+      // console.log(start, end); // eslint-disable-line no-console
```
c-w: Nit: remove commented out code.

app/server/static/components/annotator_qanda.vue
```diff
@@ -0,0 +1,179 @@
...
+        return false;
+      }
+      if (this.startOffset > this.text.length || this.endOffset > this.text.length) {
+        return false;
+      }
+      if (this.startOffset < 0 || this.endOffset < 0) {
+        return false;
+      }
+      for (let i = 0; i < this.entityPositions.length; i++) {
+        const e = this.entityPositions[i];
```
c-w: Nit: can be simplified with a [for-of loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of).

app/api/migrations/0004_question_and_answer_creation.py
```diff
@@ -0,0 +1,49 @@
+# Generated by Django 2.1.7 on 2019-07-09 18:36
```
c-w: Consider squashing the migrations into one migration which lights up the Q&A feature.

app/api/managers.py
```diff
@@ -31,3 +31,18 @@ def get_label_per_data(self, project):
             user_count[d['user__username']] += d['user__count']
 
         return label_count, user_count
+
+
+class QandAAnnotationManager(Manager):
```
c-w: The three annotation managers are now very similar. I wonder if it'll soon be worth refactoring them... just some thoughts to get a conversation going:

```py
class BaseAnnotationManager(Manager):
    pivots = ()
    aggregates = ()

    def get_label_per_data(self, project):
        label_count = Counter()
        user_count = Counter()
        docs = project.documents.all()
        annotations = self.filter(document_id__in=docs.all())

        pivot1, pivot2 = self.pivots
        agg1, agg2 = self.aggregates
        for d in annotations.values(pivot1, pivot2).annotate(Count(agg1), Count(agg2)):
            label_count[d[pivot1]] += d[f'{agg1}__count']
            user_count[d[pivot2]] += d[f'{agg2}__count']

        return label_count, user_count


class AnnotationManager(BaseAnnotationManager):
    pivots = ('label__text', 'user__username')
    aggregates = ('label', 'user')


class Seq2seqAnnotationManager(BaseAnnotationManager):
    pivots = ('text', 'user__username')
    aggregates = ('text', 'user')


class QandAAnnotationManager(BaseAnnotationManager):
    pivots = ('response', 'user__username')
    aggregates = ('response', 'user')
```

app/api/models.py
```diff
@@ -182,6 +211,7 @@ class Meta:
 
 class Document(models.Model):
     text = models.TextField()
+    extra_text = models.TextField(blank=True)
```
c-w: Wondering if instead we could/should use the `meta` field to store this extra data. To make working with `meta` easier, we could consider adding a computed property like this:

```py
@property
def metadata(self):
    return json.loads(self.meta)
```

app/server/static/package-lock.json
guillim: Agreed. However, I just git merged with master branch, and it has `"@babel/highlight": "^7.0.0"` now, so I left it

app/server/static/components/question_and_answer.vue
guillim: yep

guillim: yep

app/server/static/components/question_and_answer.vue
guillim: true

app/server/static/components/annotator_qanda.vue
guillim: that's right. Probably inherited from testing purposes. I changed it

app/server/static/components/annotator_qanda.vue
guillim: good idea. Just modified your way

app/server/static/components/annotator_qanda.vue
guillim: I am ok with renaming `e` into `entity`

app/api/models.py
guillim: From my understanding, `meta` was related to external information such as `external_id`. Here, this property is core to some task that may require more than one text. It could fit in the meta I believe, but made more sense to put it as a property to me. your call.

app/api/managers.py
guillim: I definitly agree that some factorisation could happen soon.

app/server/static/components/annotator_qanda.vue
guillim: I renamed it `emptyChunk` because that's what they are :)

app/server/static/components/annotator_qanda.vue
guillim: True

app/server/static/components/annotator_qanda.vue
guillim: Done

app/api/migrations/0004_question_and_answer_creation.py
guillim: done

guillim:
Don't know exactly why... but the new database `mssql` fails on Travis CI. It successfully passes  `sqlite` and `postgres` databases though... Any clue on how to fix this @c-w ?

icoxfog417:
[`django-pyodbc-azure==2.1.0.0` seems not be installed](https://travis-ci.com/chakki-works/doccano/jobs/223494900#L801). This requirement is [added recently](https://github.com/chakki-works/doccano/commit/47bbb741d2aeb871c323a3af9f4abdb7f51e9cc2#diff-b4ef698db8ca845e5845c4618278f29a) so you have to apply this commit to your branch.

guillim:
> [`django-pyodbc-azure==2.1.0.0` seems not be installed](https://travis-ci.com/chakki-works/doccano/jobs/223494900#L801). This requirement is [added recently](https://github.com/chakki-works/doccano/commit/47bbb741d2aeb871c323a3af9f4abdb7f51e9cc2#diff-b4ef698db8ca845e5845c4618278f29a) so you have to apply this commit to your branch.

still having the issue, any other clue ?

c-w:
@guillim The current build failure no longer is related to the environment but instead is a model error:

> django.db.utils.ProgrammingError: ('42000', "[42000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Column 'text' in table 'api_qandaannotation' is of a type that is invalid for use as a key column in an index. (1919) (SQLExecDirectW)")

I'd try to take a look at the schema created by `manage.py migrate` and see what limitations SQL server has that must be worked around.

guillim:
Ok. I finally found out what was the issue with Microsoft SQL Server: they don't accept TextField when used as a _unique_together_ property 

So I changed it to a CharField. However, it can't be larger than 500 characters, which may be a limitation in case an answer to a question is larger than 500 characters. I guess if this happens, an issue will be raised !

Thanks for the debugging tip @c-w

Issue #308, opened 2019-07-27
Deployment to Heroku fails


manrajgrover:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Heroku
- **Python version**: 3.6.8

### Describe the problem
I tried deploying to Heroku using the button in README and it failed.

### Source code / logs

```
-----> Node.js app detected
       
-----> Creating runtime environment
       
       NPM_CONFIG_LOGLEVEL=error
       NODE_ENV=production
       NODE_MODULES_CACHE=true
       NODE_VERBOSE=false
       
-----> Installing binaries
       engines.node (package.json):  8.x
       engines.npm (package.json):   unspecified (use default)
       
       Resolving node version 8.x...
       Downloading and installing node 8.16.0...
       Using default npm version: 6.4.1
       
-----> Installing dependencies
       Installing node modules (package.json)
       up to date in 0.47s
       found 0 vulnerabilities
       
       
-----> Build
       Running heroku-postbuild
       
       > @ heroku-postbuild /tmp/build_50fb938815d4a3f443e8b5b29b0f65ee
       > sh tools/heroku.sh build
       
       
       > swiper@4.3.3 postinstall /tmp/build_50fb938815d4a3f443e8b5b29b0f65ee/app/server/static/node_modules/swiper
       > node -e "console.log('\u001b[35m\u001b[1mLove Swiper? Support Vladimir\'s work by donating or pledging on patreon:\u001b[22m\u001b[39m\n > \u001b[32mhttps://patreon.com/vladimirkharlampidi\u001b[0m\n')"
       
       Love Swiper? Support Vladimir's work by donating or pledging on patreon:
        > https://patreon.com/vladimirkharlampidi
       
       added 56 packages from 328 contributors and audited 8432 packages in 7.629s
       found 125 vulnerabilities (4 moderate, 121 high)
         run `npm audit fix` to fix them, or `npm audit` for details
       
       > husky@0.13.4 install /tmp/build_50fb938815d4a3f443e8b5b29b0f65ee/app/server/static/node_modules/husky
       > node ./bin/install.js
       
       husky
       setting up hooks
       Can't find .git directory, skipping Git hooks installation
       added 909 packages from 495 contributors and audited 8432 packages in 24.255s
       found 125 vulnerabilities (4 moderate, 121 high)
         run `npm audit fix` to fix them, or `npm audit` for details
       Hash: 1eb0fe5b5d2931821759
[...]

PR #311, opened 2019-07-29
[Issue #308] Fixes heroku deployment issue

manrajgrover:
## Description

Based on fix recommended in #308, this PR attempts to fix the deployment on Heroku.

## Related issue

#308 

## People to notify
@icoxfog417

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,9 @@ RUN curl -sL "https://deb.nodesource.com/setup_${NODE_VERSION}" | bash - \
  && apt-get install --no-install-recommends -y \
       nodejs=8.16.0-1nodesource1
 
-RUN apt-get install --no-install-recommends -y \
+RUN apt-get update \
+ && apt-get install --no-install-recommends -y \
+      g++ \
       unixodbc-dev=2.3.4-1
 
 COPY app/server/static/package*.json /doccano/app/server/static/


icoxfog417:
Ordinary deployment to Heroku is done by `app.json`  which call `heroku.sh`. Of course, you can use docker container by `container:push`. Did you have trouble with `container:push`? If so, we have to fix both `heroku.sh` and `Dockerfile`.

manrajgrover:
@icoxfog417 Nope, I just tried it using the option given in REAMDE (using the button). I'm not sure of the workflow here but I think the build process takes place inside the container. Since the shell script does not contain any Ubuntu dependencies, I think the change in Dockerfile will fix this. Please correct me if I'm wrong.

icoxfog417:
Heroku button (on README) does not use a Docker container. We have to fix `heroku.sh`.

I want to integrate the deployment process to Dockerfile. `heroku.yml` and set `stack` to `container` at `app.json` will enable it. But I'm not confirmed yet.

* [Building Docker Images with heroku.yml](https://devcenter.heroku.com/articles/build-docker-images-heroku-yml#review-apps-and-app-json)
* [heroku-docker-nginx-example](https://github.com/rjoonas/heroku-docker-nginx-example)

manrajgrover:
@icoxfog417 Why are we closing this? I can make changes in `heroku.sh` and push them.

PR #317, opened 2019-08-01
Enhancement/Enable token authorization for API calls

c-w:
As discussed in https://github.com/chakki-works/doccano/issues/299, there is desire to interact with Doccano via the API since this enables integrating Doccano with others systems, e.g. fetch data from system A, push it into Doccano, and later sync the labels to system B.

In order to facilitate usage of Doccano via the API, this pull request adds token authentication to the API:

```sh
# fetch an auth token once for non-interactive usage of the API
curl 'http://localhost:8000/v1/auth-token' -d '{"username": "user", "password": "pass"}' -H 'Content-Type: application/json'

# at a later point, can use the auth token to interact with the API
curl 'http://localhost:8000/v1/me' -H 'Authorization: Token xxxxxxxxx'
```

### DIFF
--- a/app/api/urls.py
+++ b/app/api/urls.py
@@ -1,4 +1,5 @@
 from django.urls import path
+from rest_framework.authtoken.views import obtain_auth_token
 from rest_framework.urlpatterns import format_suffix_patterns
 
 from .views import Me, Features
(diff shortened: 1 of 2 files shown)


kechan:
I tried: 

`curl 'http://localhost:8000/v1/auth-token' -d '{"username": "user", "password": "pass"}' -H 'Content-Type: application/json'`

But failed, among the message: 

You are seeing this message because this site requires a CSRF cookie when submitting forms. This cookie is required for security reasons, to ensure that your browser is not being hijacked by third parties.

PR #319, opened 2019-08-01
Enable sorting of documents in frontend

BatslyAdams:
The branch adds the _ordering_ parameter to provide live sorting (created/updated ascending/descending) via a new dropdown in the sidebar.
This implementation will reset the active annotation back to the first item (offset 0) on update.

sort-order


Hironsan:
For my memo:
image

PR #333, opened 2019-08-13
Feature/roles

margaretmeehan:
This PR adds a roles system in order to manage the users' roles and permissions for Doccano projects. Collaboration with @armanrahman22 and @rbinrais. 

- Adds Role and RoleMapping models
- Adds a Project Admin, Annotator, and Annotation Approver role in settings. Users are given permission to access pages and actions based on their role. 
- Creates a Users, Roles, RoleMappingList and RoleMappingDetail API in order to edit roles in the UI.
- Adds a new Users interface so that administrators can assign roles to users within a project.

62495929-64caa600-b7a5-11e9-83f7-51679ff9a8ba

The roles are defined as follows:

View ↓ / Role →  | Project Admin  | Annotator | Annotation Approver |
---------------- |  ------------- | --------- | ------------------- |
Me               | x              | x         | x                   |
Features         | x              |x          | x                   |
ProjectList      | x              |x          | x                   | 
ProjectDetail    | x              |x          | x                   |
StatisticsAPI    | x              |x          | x                   |
ApproveLabelsAPI | x              |           | x                   |
LabelList        | x              |x          | x                   |
LabelDetail      | x              |x          | x                   |
DocumentList     | x              |x          | x                   |
DocumentDetail   | x              |x          | x                   |
AnnotationList   | x              |x          | x                   |
AnnotationDetail | x              |x          | x                   |
TextUploadAPI    | x              |           |                     |
CloudUploadAPI   | x              |           |                     |
TextDownloadAPI  | x              |           |                     |


atif-zakaria:
still not working

c-w:
@atif-zakaria Could you specify what you tried and what isn't working for you? Thanks!

atif-zakaria:
i've created all the users ans all the stuffs so i have two questions:
1/ i can't find the users icone as the PR shows 
2/lets say that we are 5 persons and we won't to work on the same project on 5 different Pcs  how we can do this !

app/server/static/pages/users.js
```diff
@@ -0,0 +1,14 @@
+import Vue from 'vue';
+import Buefy from 'buefy';
+import Users from '../components/users.vue';
+import 'buefy/dist/buefy.css';
```
Hironsan: The header color is white only in users page.
image

app/api/migrations/0004_roles.py
```diff
@@ -0,0 +1,43 @@
+# Generated by Django 2.1.7 on 2019-07-25 03:33
```
Hironsan:
```
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0004_project_collaborative_annotation, 0004_roles in api).
To fix them run 'python manage.py makemigrations --merge'
```

karlbaker02:
This is great! It'd be even better if we could add groups to projects, rather than needing to assign individual users, but that is probably a nice to have.

Are there any updates on ETA when we expect this will be ready?

app/server/static/pages/users.js
margaretmeehan: Resolved: 

users

Hironsan:
Thank you for your great work!

PR #348, opened 2019-08-22
Bugfix/Make usage of Python 3 binary explicit

c-w:
I recently tried to run doccano on a CentOS 7 system. CentOS 7's `python` binary links to python2 by default which means that even if python3 is installed via yum, all the scripts and tooling in doccano break since they reference the default `python` binary.

Making it explicit that we're requiring python3 to run the scripts fixes this class of errors and also acts as additional documentation for future developers to make it clear that doccano does not work on python2.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,7 +55,7 @@ wheels/
 MANIFEST
 
 # PyInstaller
-#  Usually these files are written by a python script from a template
+#  Usually these files are written by a Python script from a template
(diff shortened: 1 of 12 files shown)


c-w:
test

icoxfog417:
Python 2.7 is going to be the end of support at 2020/1/1 (2 months 
 later). Python3 will be default `python` at any platform in the near future.

For that reason, I think this change is unnecessary. Whenever we have to install the doccano to the default Python2 system, we can use Docker for running doccano.

PR #365, opened 2019-08-30
Fixes spaCy export

erip:
In lieu of seemingly abandoned PR #264.


erip:
@c-w any thoughts about this?

erip:
@Hironsan do you have any feedback?

app/api/utils.py
```diff
@@ -451,6 +471,67 @@ def paint(self, documents):
...
+            annotations = d.pop('annotations')
+            entities = []
+            for a in annotations:
+                label_text = label_texts[a['label']]
+                entities.append([a["start_offset"], a["end_offset"], label_text])
+            res.append((text, {"entities": entities}))
+        return [res]
+
+
+class Color:
```
c-w: This class was removed in https://github.com/doccano/doccano/pull/393. Looks like this is a merge artifact?

app/api/utils.py
```diff
@@ -385,6 +385,26 @@ def parse(self, file):
         if data:
             yield data
 
+class SpacyParser(FileParser):
+
+    def parse(self, file):
+        file = io.TextIOWrapper(file, encoding='utf-8')
```
c-w: In https://github.com/doccano/doccano/pull/399 we introduced `EncodedIO`. Please also add this here before calling `TextIOWrapper`.

app/api/utils.py
```diff
@@ -451,6 +471,67 @@ def paint(self, documents):
...
 
 
+class SpacyPainter(JSONPainter):
+
+    def paint(self, documents):
+        data = super().paint(documents)
+        res = []
+
+        label_ids = {a['label'] for d in data for a in d['annotations']}
+        label_texts = {l['id']: l['text'] for l in Label.objects.filter(id__in=label_ids).values('text', 'id')}
```
c-w: We could avoid this database lookup if in the annotation serializers ([DocumentAnnotationSerializer](https://github.com/doccano/doccano/blob/72cafaf881aa239393566278613a725c35b60ee8/app/api/serializers.py#L162) and [SequenceAnnotationSerializer](https://github.com/doccano/doccano/blob/72cafaf881aa239393566278613a725c35b60ee8/app/api/serializers.py#L173)) we serialized the label field as an object (e.g. `{"id": 1, "text": "the-label"}`) instead of just an id. This would also be beneficial for the CSV/JSONL data export where currently we're exporting labels as ids instead of as human readable text. @Hironsan what are your thoughts about changing the annotation serializers to include a text version of the label instead of only the id?

PR #388, opened 2019-10-05
[discussion] GCP Run - "Free" deployment solution for small teams

almeta-io:
We were looking for an annotation tool and decided to go with yours, Nice work btw!

Given the following aspects:
- We are a very small team (<10 people)
- Cost should be as low as possible
- We annotate rather small datasets

We decided to go with GCP Run as a deployment solution, and it worked rather smoothly (thanks for having a Dockerfile ready)!  
Now, we are practically good to go, and our endpoint is up and running, but we were wondering if you think other people would be interested, given the aspects we just mentioned.

GCP Cloud Run is designed for stateless servers:
- Has a max RAM of 2GB
- Hard disk is not persistent among different deployments (i.e. once you deploy a new version, you will lose all data you have).

To do:
- [ ] Upload a Docker image of the latest doccano/master
- [ ] Update "Deploy to Cloud Run link"

If you want to test `cloudbuild.yaml` link this repo to Cloud Build and watch the magic! \^\^

*Do not worry about uploading and hosting the Docker image, we will easily and happily do that for free.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Doccano can be deployed to AWS ([Cloudformation](https://docs.aws.amazon.com/AWS
 
 If you have a very small team, Doccano can also be deployed to GCP ([Cloud Run](https://cloud.google.com/run)) by clicking on the button below:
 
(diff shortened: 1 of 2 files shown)


almeta-io:
@Hironsan are you the one approving those?

PR #398, opened 2019-10-11
Bugfix/Enable superuser to delete user annotations

c-w:
As reported in https://github.com/chakki-works/doccano/issues/385, superusers currently can't delete annotations created by another user. This pull request enables superusers to delete annotations even if they were created by another user.

Resolves https://github.com/chakki-works/doccano/issues/385


jonhilgart22:
Bump! Can we merge this in :)

prokotg:
hey ! Thank you for this contribution! As I can see,there is only super user you consider rn. When I look at administration panel where I give users permissions (through groups) there isan option for a user to have the following permissions 
image
Should not they also be able to change other people's annotations? Or do I miss sth out here?

Cheers! :)

PR #399, opened 2019-10-14
Bugfix/Fix data import from non UTF-8 files

c-w:
As reported in https://github.com/chakki-works/doccano/issues/88, it's currently impossible to import data from files that are not encoded in UTF-8. This pull request fixes this limitation by leveraging chardet's [UniversalDetector](https://chardet.readthedocs.io/en/latest/usage.html#example-detecting-encoding-incrementally) to automatically detect the encoding of the uploaded file without buffering the entire file in memory.

Resolves https://github.com/chakki-works/doccano/issues/88


Hironsan:
Thanks!

PR #405, opened 2019-10-17
ami id, docker ports, instance type

adamkgray:
Changes:

Ami Id:
I made an issue a few weeks ago about the hard coded AMI id.
https://github.com/chakki-works/doccano/issues/397
The documentation was updated but the AMI is still hard coded.
This solution will dynmically pull the correct ubuntu 16 ami id no matter what region you are in.

Docker ports:
Not sure if I am missing something here, but the app serves on port 8000, but the user data in the cfn template maps 80 to 80. So I changed it to map 80 to 8000. It works when I launch it in my AWS account.

Instance type:
The cfn template asks the user to select an instance type. But then it used a hard coded one anyways. So I changed it to use the selected instance type. I assume this was the intended behavior? I also added t3 instance types. Depending on the region you are in, different instance types may or may not be available. It's good to allow users the choice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ git push heroku master
 
 Doccano can be deployed to AWS ([Cloudformation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/Welcome.html)) by clicking on the button below:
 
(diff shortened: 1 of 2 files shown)


BrambleXu:
@icoxfog417 
Please merge this PR. 
I cannot merge it due to no collaborator role (no merge button shows on the PR page).

PR #412, opened 2019-10-21
Auto-labeling implementation

ThibaudReal:
Implementation of a solution for auto-labeling (sequence labelling projects only) as discussed in https://github.com/chakki-works/doccano/issues/191

- Solution based on spaCy’s ner module (multi-language)

- New button that allows an admin to start auto-labeling (and removes unapproved annotations)

- For now, if a document has no annotation, it will never be described as ‘validated’. I propose  an other way to define a document as ‘validated’ : a document is ‘validated' when it has been approved by an admin. I changed this behaviour because otherwise we couldn’t differentiate documents labeled automatically from others. 

Possible improvements :

- Add a shortcut to quickly validate one document (Enter?)

- Currently, the labels suggested by the algorithm are saved as if they were done by the admin clicking on the button; so a ‘normal’ user can’t change it. It would be better to add these new labels as if they were done by an other user (named ‘computer' for example)

### DIFF
--- a/app/api/urls.py
+++ b/app/api/urls.py
@@ -8,7 +8,7 @@ from .views import LabelList, LabelDetail, ApproveLabelsAPI
 from .views import DocumentList, DocumentDetail
 from .views import AnnotationList, AnnotationDetail
 from .views import TextUploadAPI, TextDownloadAPI, CloudUploadAPI
-from .views import StatisticsAPI
+from .views import StatisticsAPI, AutoLabelAPI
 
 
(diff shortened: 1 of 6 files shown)


ThibaudReal:
Some tests have failed but it should be due to the 'remaining' bar that is now counting the number of validated documents instead of the number of annotated documents.

emilysauer:
Do I need to also install SpaCy in order for this to work, if so where would I be installing it?

PR #441, opened 2019-11-11
Enhancement/Add volume for SQLite DB in doccano container

c-w:
This pull request moves the SQLite database file used in Doccano's Dockerfile to a dedicated directory `/data` and binds this directory using a [VOLUME instruction](https://docs.docker.com/engine/reference/builder/#volume).

This enables us to persist Doccano's database file across container restarts in a filesystem/platform independent way. The README has been updated with instructions on how to leverage this functionality.

Resolves https://github.com/chakki-works/doccano/issues/438

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,14 +60,25 @@ git clone https://github.com/chakki-works/doccano.git --config core.autocrlf=inp
 
 ### Docker
 
+As a one-time setup, create a Docker container for Doccano:
+
 ```bash
-$ docker pull chakkiworks/doccano
-$ docker run -d --rm --name doccano \
-   -e "ADMIN_USERNAME=admin" \
-   -e "ADMIN_EMAIL=admin@example.com" \
-   -e "ADMIN_PASSWORD=password" \
-   -p 8000:8000 chakkiworks/doccano
+docker pull chakkiworks/doccano
+docker container create --name doccano \
+  -e "ADMIN_USERNAME=admin" \
+  -e "ADMIN_EMAIL=admin@example.com" \
+  -e "ADMIN_PASSWORD=password" \
+  -p 8000:8000 chakkiworks/doccano
 ```
+
+Next, start Doccano by running the container:
+
+```bash
+docker container start doccano
+```
+
+To stop the container, run `docker container stop doccano -t 5`.
+All data created in the container will persist across restarts.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -136,16 +136,25 @@ Depending on your installation method, there are three options:
...
-First, run a Docker container:
+As a one-time setup, create a Docker container for Doccano:
 
 ```bash
-docker run -d --rm --name doccano \
+docker container create --name doccano \
   -e "ADMIN_USERNAME=admin" \
   -e "ADMIN_EMAIL=admin@example.com" \
   -e "ADMIN_PASSWORD=password" \
   -p 8000:8000 chakkiworks/doccano
```
louisguitton: Love this feature. Haven't tested the code but had a question at first read: with `docker container create` one does not need to bind the volume as with `docker run` ?
The `-v` flag is documented for `docker container create` [here](https://docs.docker.com/engine/reference/commandline/container_create/).
And I would have assumed the new command would mention the container like so maybe:
```sh
docker container create --name doccano \
  -v ./data:/data
  -e "ADMIN_USERNAME=admin" \
  -e "ADMIN_EMAIL=admin@example.com" \
  -e "ADMIN_PASSWORD=password" \
  -p 8000:8000 chakkiworks/doccano
```

c-w: Thanks for the feedback. Using a host folder mount (`-v ./data:/data`) can have gotchas on some platforms, e.g. on Windows if the drive isn't shared with the Docker daemon, on Linux if the user in the Docker process doesn't have permissions on the host system, etc. and requires the user to be more familiar with Docker to troubleshoot issues.

To ensure that Doccano can work easily on all systems, this pull request currently uses a [VOLUME instruction](https://docs.docker.com/engine/reference/builder/#volume) in the Dockerfile which creates and binds a Docker volume to the container, similar to what `docker container create -v $(docker volume create):/data ...` would do.

Do you have any concerns with this approach?

louisguitton: None at all. I was not that clear about that VOLUME instruction. Thanks for the link.

orico:
did you merge this yet?

README.md
orico: I can confirm that -v is needed in order to preserve the data. as far as i understand the volume instruction still writes it internally.

Issue #95, opened 2019-02-22
WIP : Annotation of audios for Speech-to-Text parallel data set building 


tayciryahmed:
### Describe the problem
Given the relative scarcity of tools for building a speech-to-text dataset (most standard tools are desktop-based now), it would be great if doccano could serve that purpose. At its core, I believe that would mean each audio is an input base64 string (displayed in the interface as an audio html5), then there would a form below (like in a Translation project) that allow you to enter annotations. 

The main functionality that would be tricky would likely be displaying the audios in the interface (easy using html5) and importing the audio data to doccano (we can represent audio files in Base64 encoding and load them as csv or json).

Does this sound doable?

tayciryahmed:
If you think that is aligned with your vision of the project, I can work on it and submit an MR.

Hironsan:
As I showed in [feature ideas](https://github.com/chakki-works/doccano/projects/2#card-16405643), I think it is important for users to increase the number of tasks they can annotate.

So, if you can work on the audio annotation feature and submit a PR, we are very happy!

BrambleXu:
@tayciryahmed The model has changed a lot #110. If you are working on the audio annotation feature, please update to the latest version of doccano.

PR #121, opened 2019-03-24
Annotation of audios for Speech-to-Text parallel data set building

tayciryahmed:
Solves #95. 

Below is the tracked task list:

- [x] Load audio
- [x] Save audio (in the future maybe, encode the audio as base64 and save nothing in the disk). 
- [x] Annotation interface 
- [x] Save annotations 
- [x] Fix previous / next annotation buttons 
- [x] Export annotations
- [x] Improve the styling (annotation interface and upload interface)
- [x] Finish tests (add tests for upload audio + annotation integrity)
- [x] Clean code and remove unnecessary blocks

PR #495, opened 2019-12-12
Feature/Speech to text transcription

c-w:
This pull request is based on the work of @tayciryahmed in https://github.com/doccano/doccano/pull/121 and implements speech-to-text transcription in Doccano.

To keep things simple, the implementation for now uses [html5 audio](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio) instead of something more sophisticated like [wavesurfer](https://wavesurfer-js.org/). This can be improved in the future if the requirement arises. The `alt+p` keyboard shortcut has been introduced to play/pause the audio player.

Animation showing speech to text transcription

For ease-of-use, speech-to-text data can be imported either by posting audio files (MP3, WAV, etc.) or by uploading a JSONL manifest that encodes the audio as [data URIs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) or URLs to the audio files.

To make it easier to identify and distinguish audio files, the document left-navigation has been updated to display a file name (instead of file content) if the `meta.filename` attribute is set.

Resolves https://github.com/doccano/doccano/issues/95


harmw:
nice!

First question that comes to mind is (glimpsed through the code - couldn't find it), does this allow for pre-transcribed text to be added as part of the upload?
I'm thinking of having a pipeline that does the transcription, and we can utilise doccano to do the (human-in-the-loop) validation/addition of what our models came up with (a visual representation of the P-value would probably be part of that too).

Issue #560, opened 2020-02-11
LDAP-Support


scheiblr:
Feature description
---------
It would be nice to be able to connect doccano to an LDAP-server for user and group management.

c-w:
You may be interested in https://github.com/doccano/doccano/pull/75 and https://github.com/doccano/doccano/pull/210 which added support for authentication via Azure Active Directory and Github as well as management of the admin status via those systems. Adding support for additional external identity providers should be fairly straight forward by extending this prior work as long as the provider is implemented in [Python Social Auth](https://python-social-auth.readthedocs.io/en/latest/backends/index.html#social-backends).

armandleopold:
UP

armandleopold:
UP

PR #586, opened 2020-03-03
Feature/LDAP support

vallejo-demos:
This pull request is about the [requested feature to be able to connect doccano to an LDAP server](https://github.com/doccano/doccano/issues/560).

LDAP support for doccano is for many organizations a very important feature since they already have an elaborated LDAP authentication infrastructure. Also group management is often included in their existing infrastructure.

The provided solution integrates seamlessly into doccano. Only another Django authentication backend and its configuration are needed. The configuration for LDAP should be part of the existing environment variables. Therefore it integrates well with the existing flow.

Since the LDAP authentication backend integrates well into Django applications, the existing login page of doccano doesn't need to be changed. Everything should work fine.

### DIFF
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -14,6 +14,8 @@ from os import path
 
 import django_heroku
 import dj_database_url
+import ldap
+from django_auth_ldap.config import LDAPSearch
 from environs import Env
 from furl import furl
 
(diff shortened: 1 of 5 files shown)


vallejo-demos:
Hi Clemens and thanks for taking time to review this request.

You are absolutely right, it would be helpful to add the required installation commands to the Doccano dockerfile.

`apt-get install ldap-utils` should do the trick. I hope I put it in the right place...

tools/dev-django.sh
```diff
@@ -15,6 +15,7 @@ fi
 
 echo "Installing dependencies"
 apt-get update && apt-get install -y g++ unixodbc-dev # pyodbc build dependencies
+apt-get install -y ldap-utils # ldap dependencies
```
c-w: `dev-django.sh` is only used for running at development time. For any dependencies that are required in production (which I assume LDAP utils falls under), consider following the same pattern that we have for the MS SQL dependencies:

https://github.com/doccano/doccano/blob/dce6f035a2ea93aedc1ada064e68effdd04420ed/Dockerfile#L46-L47

scheiblr: Any progress on this?

scheiblr:
Any news on that? Do you plan to integrate this feature soon?

c-w:
@scheiblr Could you rebase the pull request with latest master to resolve the conflicts and also make sure that the TravisCI build passes?

flavienbwk:
Was this project abandoned @vallejo-demos ? It is great !

armandleopold:
Up

juguerin:
Hi, we started to use Doccano in our institute a couple of months ago and now we have to find a way to use it more widely. The LDAP and access groups feature would be more than required. What's the feature status now?

PR #603, opened 2020-03-10
Enable to use digits as shortcut keys and resolve #599

### LINKED ISSUE #599: [Enhancement request] Allow digit keys as shortkeys
Feature description
---------
English letters are allowed as shortkeys for annotation now, only.

Proposition: allow English letters and digits as shortkeys.

Hironsan:


kuraga:
@Hironsan , I see digit's in possible values list in Django Admin but there are no digits in "Create a label" UI dialog.

PR #612, opened 2020-03-12
Adding the missing file webpack-stats.json

sivaprakash123:
This file is not generated during we run npm install. Hope it should in the repo. 

Thanks,
Sivaprakash R


divankovic:
Yes, this file is required, the development mode setup does not work as described in the README. Also you should go to localhost:8000, not localhost:3000.

PR #675, opened 2020-04-08
Add Okta support

### LINKED ISSUE #674: Add Okta Integration for OAuth
Feature description
---------

Add Okta OpenID and OAuth support as other social login options.

setu4993:
First time-contributor and working with OAuth / Django. I took inspiration from #75 but I probably missed some things. Would appreciate some help on the tests to add, and missing changes required.
Resolves #674.


setu4993:
Hi @Hironsan, @c-w, rest of the team. Please review this when you have a chance.

setu4993:
Hello @Hironsan, @c-w, I think this PR is now ready for review. I have added everything that I could see and tested it using an Okta domain.

setu4993:
Hi @Hironsan, @c-w, @icoxfog417, please review this when you have a chance.

The code quality failing CI is from the test where it assumes there's a hard-coded password or secret.

app/server/tests/test_social_auth.py
```diff
@@ -93,3 +95,75 @@ def test_fetch_permissions_not_admin(self):
...
+            user=user,
+            backend=self.backend,
+            response={'access_token': self.access_token},
+        )
+
+        self.assertFalse(user.is_superuser)
+
+
+@override_settings(SOCIAL_AUTH_OKTA_OPENIDCONNECT_KEY='0000000000aaaaaaaaaa')
+@override_settings(SOCIAL_AUTH_OKTA_OPENIDCONNECT_SECRET='bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb=')
```
c-w, line 137: You should be able to suppress the codacy false positive by adding `# nosec` to this line (see [docs](https://github.com/PyCQA/bandit#exclusions)).

app/requirements.txt
```diff
@@ -29,13 +29,15 @@ model-mommy==1.6.0
...
 pyexcel-xlsx==0.5.7
+pyjwt>=1.7.1
 python-dateutil==2.7.3
+python-jose>=3.0.0
 pytz==2018.4
 requests==2.21.0
 six==1.11.0
 seqeval==0.0.6
 social-auth-app-django==3.1.0
-social-auth-core[azuread]==3.0.0
```
c-w, line 38: Why remove the `azuread` extra dependency here? This will break users who are currently using this authorization provider.

app/requirements.txt
```diff
@@ -29,13 +29,15 @@ model-mommy==1.6.0
 psycopg2-binary==2.7.7
 pyexcel==0.5.14
 pyexcel-xlsx==0.5.7
+pyjwt>=1.7.1
 python-dateutil==2.7.3
+python-jose>=3.0.0
```
c-w, line 34: Instead of adding the `pyjwt` and `python-jose` dependencies explicitly, consider instead adding the `openidconnect` extra dependency for `social-auth-core` which will include these dependencies (see [requirements-openidconnect.txt](https://github.com/python-social-auth/social-core/blob/9d93069564a60495e0ebd697b33e16fcff14195b/requirements-openidconnect.txt)) and also will ensure that the dependencies stay in tune with any changes that the social auth project makes in the future.

app/server/social_auth.py
```diff
@@ -68,3 +70,64 @@ def fetch_azuread_permissions(strategy, details, user=None, is_new=False, *args,
...
+    is_superuser = admin_group_name in response.get("groups", [])
+    is_staff = admin_group_name in response.get("groups", [])
+
+    if user.is_superuser != is_superuser:
+        user.is_superuser = is_superuser
+        user.save()
+
+    if user.is_staff != is_staff:
+        user.is_staff = is_staff
+        user.save()
```
c-w, line 105: I would suggest to batch the two `user.save()` calls so that we avoid potential weird states with the model being half updated, perhaps something like this:

```suggestion
    user_changed = False

    if user.is_superuser != is_superuser:
        user.is_superuser = is_superuser
        user_changed = True

    if user.is_staff != is_staff:
        user.is_staff = is_staff
        user_changed = True

    if user_changed:
        user.save()
```

c-w: Same comment for `fetch_okta_openidconnect_permissions` below.

app/requirements.txt
setu4993: My bad. I (incorrectly) assumed that replacing it with `social-auth-core` would mean all the dependencies would get added including those for Azure and for OpenID. Fixing it now.

app/server/social_auth.py
setu4993: Good point, done.

app/server/tests/test_social_auth.py
setu4993: Didn't know that, nice. Thanks!

setu4993:
Thanks for the review and feedback, @c-w. I've updated the PR based on your comments, and the CI build is also passing.

app/requirements.txt
```diff
@@ -34,8 +34,9 @@ pytz==2018.4
 requests==2.21.0
 six==1.11.0
 seqeval==0.0.6
-social-auth-app-django==3.1.0
-social-auth-core[azuread]==3.0.0
+social-auth-app-django==4.0.0
+social-auth-core[azuread]==3.3.3
+social-auth-core[openidconnect]==3.3.3
```
c-w, line 39: To my understanding, the canonical way to specify multiple extra dependencies would be like this:

```suggestion
social-auth-core[azuread,openidconnect]==3.3.3
```

setu4993: Updated.

Hironsan:
Is this WIP or finished?

setu4993:
@Hironsan : Hi, this is finished and ready for review.

Issue #584, opened 2020-03-02
Documents jump out of order in sequence annotation project


mmisiewicz-yext:
How to reproduce the behaviour
---------
When clicking through documents to label, or using the arrow keys, sometimes documents jump out of order. The page indicator on the upper right corner only increments by one item at a time. 

[Here is a video that might make this more clear](https://michaelmisiewicz.com/stuff/doccano_pagination_bug_demo.mov). The bug first occurs at 0:08 and I am able to cause it again a few times. 

There are no obvious errors in the browser console or logs from Nginx or Django that appear to explain it.

Behavior appears to occur both in Safari and Chrome on Mac. I have observed it occurring around the transition from 10-11 which makes it seem like a pagination bug, but possibly it occurs elsewhere in sequence of documents (e.g. 145 to 146) as well. 

Your Environment
---------
*   Operating System:
CentOS 7.5 on AWS, running Docker 18.06.1-ce, build e68fc7a
*   Python Version Used:
According to `Dockerfile`, 3.6.
*   When you install doccano:
Today, from HEAD
*   How did you install doccano (Heroku button etc):
Docker compose. i.e.: `docker-compose -f docker-compose.prod.yml up -d`.

icoxfog417:
related #555.

can-keklik:
There can be another clue. In list view (rows per page=10) last element in each page is replaced with a certain document. That document can be seen at the end of each page. This also occurs in annotation view. 

When you increase rows per page to 50, you can see that 10th, 20th, 30th and 40th elements are shown correctly, yet it is not possible to annotate them. Clicking on annotate for them takes you to repeating document. And repeating document still stays at the end of the list page. So when you are at 31th document, it means you have skipped 3 of them like:

1, 2, 3, ... 9, Repeating doc, 11, 12 ,13, ..., 19, Repeating doc, 21, 22, 23, ..., 29, Repeating doc, 31

BTW, I use shuffling for this project.

Issue #555, opened 2020-02-06
Scroll multiple times on button press


nbroad1881:
When using the arrow keys to go to the next sample, occasionally it flips through multiple samples very quickly.

See attached gif

How to reproduce the behaviour
---------
I was annotating using keyboard shortcuts and changing samples with left/right arrow keys.

Your Environment
---------
Mac OS Catalina 10.15.2 (19C57)
Docker version 19.03.5, build 633a0ea
Python 2.7.16
Installed with docker compose
![doccano_gif](https://user-images.githubusercontent.com/24982805/73976634-b68bee80-48f6-11ea-9ae7-2c8ab756ffdc.gif)

icoxfog417:
Let me confirm the behavior you faced. If you press the left/right key, doccano occasionally skips some samples, ok?

The #488 maybe related issue.  Would you please update the doccano and confirm the phenomenon?

icoxfog417:
related: #502

nbroad1881:
> Let me confirm the behavior you faced. If you press the left/right key, doccano occasionally skips some samples, ok?
Correct
 
> The #488 maybe related issue. Would you please update the doccano and confirm the phenomenon?
What do you mean by update?

icoxfog417:
Thank you for the response. The "update" means to use the latest doccano / docker-compose if you use the older one.

Hironsan:
In fact, doccano fetches 10 documents at one time and stores them in local memory. If we press the left/right key, doccano changes the current document index(0-9). If we press the right key when the current document index is 9, doccano changes it to 0 and fetches new 10 documents. Due to the network delay, there is some time to show old documents.

kuraga:
> Due to the network delay, there is some time to show old documents.

So you confirm it is a bug?

aam1r:
This is a bad bug. I have people on my team who are skipping over dozens of documents because of this.

They have to consistently go back and find "Undone" tasks and re-do them and it's slowing us down.

Can we get a fix for this please?

PR #722, opened 2020-04-29
iss555: awaiting document loading before displaying text

Citronelol:
Hi there,

First of all, I want to thank you warmly for all the work around doccano, this is an awesome project!

I had the same issue than #555 and #584 and wanted to see how to fix it. It appears indeed this is due to a delay when getting the new batch of documents, after reaching the 10 first documents cached in memory. 

I have made the document display dependent on the `state.loading` variable, making sure we do not display anything as long as the API call has not returned.

I don't know if you wanted to fix this issue this way ? Feel free to send feedback, I'll be happy to contribute further.


Hironsan:
Thanks! It works fine.

Issue #723, opened 2020-04-30
More comprehensive error message when login fails due to active django admin session


TobiasRoeddiger:
Thank you very much for the great tool.

I spent a significant amount of time wondering why adding users and changing the admin password does not work. Whenever there is an active django admin session, logging in fails with "Incorrect username or password." 

I'd suggest changing this, e.g., with a reference to the admin page and the suggestion to logout whenever there is an ongoing admin session in the background.

Tobias

icoxfog417:
Thank you for reporting the difficulty you faced. Could you please tell us about the reproducing procedure? 

1. You use Django admin site to add users or change passwords.
2. After that. you can't log in doccano unless you close the session of the Django admin site (close page or tab).

Are the above procedures correct?

mmisiewicz-yext:
I have been able to reproduce this behavior by logging into the admin, and then attempting to log into the site via the /auth page. It seems that simply being logged in as the admin causes the error `Forbidden: /v1/auth-token` to appear in the logs. A procedure I can use to reproduce is:

1. log in to /admin
2. go to home
3. press sign in button
4. attempt to log in as same user, will fail with auth-token error. 

Logging out of the admin interface resolves the issue.

TobiasRoeddiger:
Yes, @icoxfog417 that's correct, sorry for being a bit unclear in my initial description. @mmisiewicz-yext thanks for the more detailed steps to reproduce the problem.

icoxfog417:
Thank you for the detail procedure! It seems to be difficult to solve because this behavior is caused by Django authentication system.  At first I update FAQ.

kuraga:
> At first I update FAQ

:+1: 

@TobiasRoeddiger , also related: #545, #613, #667.

mlussati:
How to solved status 500

PR #729, opened 2020-05-02
Update FAQ (#723)

icoxfog417:
Add a notification to logout from the Django admin site before login to doccano.


kuraga:
@icoxfog417 , but why in `I want to add annotators` part? It's useful before each login.

icoxfog417:
Because users open Django admin site to add annotators. The notification should be located in the situation that mainly occurs.

Issue #738, opened 2020-05-04
[Enhancement request] Avoid "That username is already taken"


kuraga:
Enhancement description
---------

I have these errors in log on each start:

```
backend_1   | Initializing database
backend_1   | Operations to perform:
backend_1   |   Apply all migrations: admin, api, auth, authtoken, contenttypes, sessions, social_django
backend_1   | Running migrations:
backend_1   |   No migrations to apply.
backend_1   |   Your models have changes that are not yet reflected in a migration, and so won't be applied.
backend_1   |   Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
backend_1   | CommandError: Error: That username is already taken.
```

Analysis
-----------

The error
```
backend_1   | CommandError: Error: That username is already taken.
```
comes from here: 
https://github.com/doccano/doccano/blob/bdba2961bf7942eb529a1591a7499b5d15af73bf/app/server/management/commands/create_admin.py#L22
i.e. here: 
https://github.com/django/django/blob/2.2.10/django/contrib/auth/management/commands/createsuperuser.py#L144-L146

Note 
https://github.com/doccano/doccano/blob/bdba2961bf7942eb529a1591a7499b5d15af73bf/app/server/management/commands/create_admin.py#L23
*doesn't* catch this `CommandError` and `handle`'s execution *is interrupted*.

icoxfog417:
solved #787

adhamhassan99:
How did you solve the issue i have the same error please help.

PR #787, opened 2020-05-19
Enhancement/avoid username is taken

cgill95:
catching the exception that results in the error message:
`backend_1   | CommandError: Error: That username is already taken.`
this message was displayed at every startup and is now caught by the try catch block.

enhancement was requested in issue #738

### DIFF
--- a/app/server/management/commands/create_admin.py
+++ b/app/server/management/commands/create_admin.py
@@ -20,10 +20,9 @@ class Command(createsuperuser.Command):
 
         try:
             super().handle(*args, **options)
-        except IntegrityError:
+            self.stderr.write(f'try executed')
+        except (IntegrityError, CommandError) as e:
             self.stderr.write(f'User {username} already exists.')
-        except CommandError:
-            pass
 
         if password:
             database = options.get('database')
--- a/app/server/management/commands/create_admin.py
+++ b/app/server/management/commands/create_admin.py
@@ -20,8 +20,11 @@ class Command(createsuperuser.Command):
 
         try:
             super().handle(*args, **options)
-        except (IntegrityError, CommandError):
-            self.stderr.write(f'User {username} already exists.')
+        except Exception as err:
+            if 'is already taken' in str(err):
+                self.stderr.write(f'User {username} already exists.')
+            else:
+                raise
 
         if password:
             database = options.get('database')
--- a/app/server/management/commands/create_admin.py
+++ b/app/server/management/commands/create_admin.py
@@ -21,7 +21,7 @@ class Command(createsuperuser.Command):
         try:
             super().handle(*args, **options)
             self.stderr.write(f'try executed')
-        except (IntegrityError, CommandError) as e:
+        except (IntegrityError, CommandError):
             pass
 
         if password:
--- a/app/server/management/commands/create_admin.py
+++ b/app/server/management/commands/create_admin.py
@@ -22,6 +22,8 @@ class Command(createsuperuser.Command):
             super().handle(*args, **options)
         except IntegrityError:
             self.stderr.write(f'User {username} already exists.')
+        except CommandError:
+            pass
 
         if password:
             database = options.get('database')
--- a/app/server/management/commands/create_admin.py
+++ b/app/server/management/commands/create_admin.py
@@ -22,7 +22,7 @@ class Command(createsuperuser.Command):
             super().handle(*args, **options)
             self.stderr.write(f'try executed')
         except (IntegrityError, CommandError) as e:
-            self.stderr.write(f'User {username} already exists.')
+            pass
 
         if password:
             database = options.get('database')
--- a/app/server/management/commands/create_admin.py
+++ b/app/server/management/commands/create_admin.py
@@ -20,13 +20,13 @@ class Command(createsuperuser.Command):
 
         try:
             super().handle(*args, **options)
-            self.stderr.write(f'try executed')
         except (IntegrityError, CommandError):
-            pass
+            self.stderr.write(f'User {username} already exists.')
 
         if password:
             database = options.get('database')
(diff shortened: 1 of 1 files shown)


kuraga:
Did you mean #738 ?

Thanks! But I don't think it's not a correct patch:
1. With your patch, `IntegrityError` exception proceeds `User {username} already exists`, `CommandError` exception proceeds nothing. But for the case of existence to use there is `CommandError` (but not only for it - should we analyze the given error more?), and the `IntegrityError` is responsible for just something else.
2. What should we do if user exists? Why `pass`/`print` only? I expect for user's password change in this case; or abort with en error.

(And I don't know correct way - that's why I haven't posted a patch.)

cgill95:
> Did you mean #738 ?

yes thats right! thanks, i fixed the link.

> Thanks! But I don't think it's not a correct patch:
> 
> 1. With your patch, `IntegrityError` exception proceeds `User {username} already exists`, `CommandError` exception proceeds nothing. But for the case of existence to use there is `CommandError` (but not only for it - should we analyze the given error more?), and the `IntegrityError` is responsible for just something else.

I did not wanna remove the IntegrityError since I never have seen that error come up on my console when starting a fresh or existing container of doccano. It was always just the CommandError that came up like you described it in your issue.

> 2. What should we do if user exists? Why `pass`/`print` only? I expect for user's password change in this case; or abort with en error.
> 
> (And I don't know correct way - that's why I haven't posted a patch.)

The IntegrityError seems to be designed for the case that the username already exists, which is always going to pop up since everytime the container is restarted the admin account is tried to be created again. To avoid that the except could just yield a `pass` instead of the `print` that it currently does.

As far as I checked it out the account is getting a new password when you change the password inside the docker-compose. If you were to change the admin password via the web interface it would get reset once you rerun the docker-compose.

cgill95:
As far as I understand it either the commit https://github.com/doccano/doccano/pull/787/commits/2581f8ab4d9abbeb8280a0ceb4d403060bf365c1 or https://github.com/doccano/doccano/pull/787/commits/86252aa314a9846a8034619914f79aa5bda79a83 could be pulled into master.

The first commit has the exceptions captured together and still prints out the error but this time stating the actual username. 

The second commit just passes over the exceptions which would remove the error entirely from the logs when starting up the containers via docker-compose. This would avoid the line coming up and thus close the issue #738 

Personally I prefer the second option since I do not see a point in logging the line on every restart of the containers, since there is no value to it.

kuraga:
> As far as I checked it out the account is getting a new password when you change the password inside the docker-compose.

Hm... AFAIK we can't **re**set password this way.

1. IMHO we should `print` everything we do.
2. IMHO we should `print` `error` we got.
3. IMHO we should `print` (or `print` and abort) here something like `Can't reset password. <Some instruction>`. (Or, reset the password.)

cgill95:
> Hm... AFAIK we can't **re**set password this way.

I tested out multiple scenarios now and the password always gets overriden even if the user already exists. I added logging for this since you made valid points that everything should be logged. So now both scenarios ((re)setting password and user already exists) get logged and the CommandError is caught with our own message to stderr

kuraga:
> > Hm... AFAIK we can't reset password this way.
>
> I tested out multiple scenarios now and the password always gets overriden even if the user already exists. 

Aha, **with your patch**. Then ok.

app/server/management/commands/create_admin.py
```diff
@@ -20,12 +20,13 @@ def handle(self, *args, **options):
 
         try:
             super().handle(*args, **options)
-        except IntegrityError:
+        except (IntegrityError, CommandError):
             self.stderr.write(f'User {username} already exists.')
```
kuraga, line 24: Could `(IntegrityError, CommandError)` exception mean something different user already exists? Should we analyze exception more (not only type/class)?

cgill95: Since the entire method is just used to create the superuser I do not think it can mean anything different as far as I understand the django documentation.

kuraga: No. Exception is [raised](https://github.com/django/django/blob/2.2.10/django/contrib/auth/management/commands/createsuperuser.py#L63-L169) in many cases.

Even `CommandError` is raised in many cases:
* if [`username == ''`](https://github.com/django/django/blob/2.2.10/django/contrib/auth/management/commands/createsuperuser.py#L88-L89) or [`username is None`](https://github.com/django/django/blob/2.2.10/django/contrib/auth/management/commands/createsuperuser.py#L141-L142),
* [here](https://github.com/django/django/blob/2.2.10/django/contrib/auth/management/commands/createsuperuser.py#L153-L154),
* if [this function](https://github.com/django/django/blob/2.2.10/django/contrib/auth/management/commands/createsuperuser.py#L197-L211) returned other than `None` (**three** cases),
* [on `ValidationError`](https://github.com/django/django/blob/2.2.10/django/contrib/auth/management/commands/createsuperuser.py#L162-L163).

That why I propose to `print(err)` only or similar.
However I don't understand how to understand if error was fatal.

And confirm: I also don't see `IntegrityError`.

kuraga, line 24: > However I don't understand how to determine if error was fatal.

> I also don't see `IntegrityError`.

I propose:

```suggestion
        except Exception as err:
            if 'is already taken' in str(err):
                self.stderr.write(f'User {username} already exists.')
            else:
                raise
```

cgill95:
Your second suggestion looked very good. I agree that I missed a few exceptions, though I did not consider anyone actually removing the admin account from the compose-file

kuraga:
:+1:

kuraga:
:+1:

kuraga:
@Hironsan , @icoxfog417 , @c-w , ping.

kuraga:
:tada:

PR #791, opened 2020-05-22
Remove legacy endpoint

### LINKED ISSUE #789: Dummy data shown when API is unreachable
---------
Build the frontend: `docker build -t doccano-frontend frontend`

Launch: `docker run --rm -it -p 3000:3000 doccano-frontend`

Browsing to `http://localhost:3000` shows the frontend, everything fine there, trying to login with fake credentials (username: `123`, password: `123`) works instead of failing with auth errors. So somewhere in the frontend container dummy data is stored and now sent to the client.
The page at `http://localhost:3000/projects` shows a list of (dummy/testing) projects, where one would expect to see nothing (ie. auth errors)

As an example, this is the login call my browser tries:

```bash
curl 'http://localhost:3000/v1/auth-token' \
-XPOST \
-H 'Content-Type: application/json;charset=utf-8' \
--data-binary '{"username":"123","password":"123"}'
```
Which returns html...

When I point (hack) the `nuxt` code to some (dummy) url (`http://localhost:12345/v1`) to use as API, it does return _Incorrect username or password_ when trying to login. 

I haven't looked at the code yet, figured I'd get this reported first.

Your Environment
---------
*   Operating System: docker/mac
*   How did you install doccano (Heroku button etc): `docker build`

harmw:
As having this configured without a working backend will return dummy
data (projects) to the browser.

Fixes #789

### DIFF
--- a/frontend/nuxt.config.js
+++ b/frontend/nuxt.config.js
@@ -26,10 +26,6 @@ export default {
     ]
   },
 
-  serverMiddleware: [
-    '~/api/index.js'
-  ],
-
   server: {
     host: '0.0.0.0' // default: localhost
   },


kuraga:
But does it prevent  log in with fake credentials?

harmw:
ah no, missed that part of it 🤦‍♂️ let me see where/why that breaks

PR #792, opened 2020-05-22
Docker components

### LINKED ISSUE #783: Proper tagging of docker containers
Feature description
---------
Looking at using this in production, I came across https://hub.docker.com/r/doccano/doccano/tags which shows only a single tag getting pushed: `latest`
This makes deployment a bit tricky, as I cannot differentiate between different version of the project etc.

The idea here would be to introduce properly tagged containers, to reflect a specific point-in-time release of the project (as per https://github.com/doccano/doccano/projects ?). And keeping `latest` as it is being used today.

Thoughts? (let me see if I can help around with this)

harmw:
Trying to resolve #783 

This introduces separation of the core _doccano_ components into two containers: `doccano-app` and `doccano-frontend`. Having these distinct artefacts will make it easier to run updates on either one of these, and makes it (a lot) easier to deploy to some orchestration platform (think: K8S).

It did mean introducing a little change on how to interact with the backend API, which means all backend traffic will be handled by the frontend at `/v1/` (using a nuxt-axios-proxy). This was required to have the backend endpoint be at a dynamic location at runtime, which will ease adoption in larger platforms.


kuraga:
Wow

frontend/Dockerfile
```diff
@@ -1,10 +1,15 @@
...
+ENV NUXT_PORT 3000
+
+CMD ["/usr/local/bin/npm", "start"]
+
 WORKDIR /app
 
 COPY . /app/
 
-RUN apk add -U git python3 make g++ \
+RUN LAYER=build \
```
Hironsan, line 11: Remove `Layer=build` to pass Codacy.

frontend/Dockerfile
```diff
@@ -1,10 +1,15 @@
 FROM node:13.2-alpine
 
+ENV NUXT_PORT 3000
+
+CMD ["/usr/local/bin/npm", "start"]
```
Hironsan, line 5: In `docker-compose.prod.yml`, it is necessary to run `npm start`.

app/Dockerfile
```diff
@@ -1,17 +1,20 @@
...
 COPY . /app/
 
 RUN apk add -U bash python3 python3-dev libpq postgresql-dev unixodbc-dev musl-dev g++ libffi-dev \
   && pip3 install --upgrade pip setuptools \
   && pip3 install --no-cache-dir -r requirements.txt \
   && ln -s /usr/bin/python3 /usr/bin/python \
-  && apk del python3-dev postgresql-dev unixodbc-dev musl-dev g++ libffi-dev
+  && apk del python3-dev postgresql-dev unixodbc-dev musl-dev g++ libffi-dev \
+  && mkdir -p /app/staticfiles \
+  && chmod 777 /app/staticfiles
```
Hironsan, line 19: Really need 777?

frontend/Dockerfile
harmw: Not sure what you're referring to :) For prod we're just going with the `CMD` that is in this `Dockerfile`, for the dev compose file we're launching `/src/frontend/dev-nuxt.sh`.

app/Dockerfile
harmw: Yes, as in prod we run these components as non-root user :)

```
backend_1   | PermissionError: [Errno 13] Permission denied: '/app/staticfiles/admin/css/vendor/select2/select2.min.d44571114a90.css'
```

Going with 777 is a bit too much, yes, but it should do for now.

```
backend_1   | Making staticfiles
```

Why is this even needed in a backend (granted, I didn't bother yet to dive into it)?

PR #800, opened 2020-05-28
chore: update documentation link

howiezhao:
Change documentation link to GitHub Pages, whether to add documentation link in the README file?

In addition, the images in the section about "AWS HTTPS settings" in the documentation are not accessible.

And also, the last of the three images at the bottom of the homepage is no longer accessible.

It looks like the version deployed on Heroku is not updated, it still use the old repository link.

### DIFF
--- a/app/server/templates/index.html
+++ b/app/server/templates/index.html
@@ -173,7 +173,7 @@
         </h2>
         <ul>
           <li>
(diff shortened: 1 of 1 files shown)


Hironsan:
Thanks. But in the commit message, please use the present tense and the first letter must be an upper case.

Issue #280, opened 2019-07-09
Annotation status on pre annotated data and status management


janostik:
At the moment the system does not support auto labelling, but we're trying to work around the issue by importing dataset which already contains predicted annotations. Unfortunately all entries in this dataset are marked as completed. That means it's impossible for annotator to start the work and see his progress.

I think in use cases like this it would make sense to mark the item as completed only if it contains annotations AND was reviewed by user (opened and user clicked next). In this use case it might be useful to have a button to "skip" entry in dataset. Not to mark item as completed accidentally.

Hironsan:
Thank you for your comment.

Currently, doccano supports "Approve Labels" feature in #258 by @c-w. It enables admin users to approve labels. So, we may be able to implement your feature by extending it. 

Idea:

* Currently, the User model and Document model have a one to many relationships.
* By extending it to many to many, a document can have multiple approval information.
* Using it, doccano will control annotation status per user.

janostik:
I guess you know best about a way to implement this feature, but I would be worried about changing the relation between user and document model. Since this will probably affect not only the status of document, but also statistics page. How would you differentiate document which has been approved by 1 person or multiple people?

At the moment what I like about the tool is simplicity. From experience the annotators are usually knowledgeable in their domain but not that skilful in technical terms. So presenting them with multiple buttons, 1 to mark item as finished (which they need to click all the time) and another one which is "next" might cause some confusion.

AI-ITVT:
I would like to back the feature request. 
Due to limited annotator resources we practically always try to create data-sets filled with predicted pre-annotations. This tends to significantly speed up the actual annotation process. I quite like doccano but using it in this scenario is next to impossible, since when all entries are marked as completed you never know which entries are really - manually - annotated,

icoxfog417:
Discussion point.

Annotation status may be...

* Not annotated
* Annotated
  * but need confirmation (bookmark?)
  * but not approved

buriy:
Also "automatically annotated" and "manually annotated" are different states.

buriy:
So if the process is:
(doc received by annotator) -> check up the markup -> set correct or incorrect -> 
(doc received by verifier) -> check up the markup -> set correct or incorrect -> done.

Then the list of possible states should be:
 - no annotations / has annotations (if anyone still wants this -- you know it's flawed when docs with no annotations are correct, e.g. for NER, or sequence labeling)
 - automatically annotated / not annotated (a filter for annotator input)
[...]

PR #818, opened 2020-06-04
Show  thumbs up icon for doc approved status on the left side view.

buriy:
- Display "Thumbs up" icon if the doc is approved.
 - Else, display "Check mark" if the doc has annotations.
 - Otherwise, display no icon.

See #280 for discussion.


buriy:
Also for the "old" frontend only.

PR #855, opened 2020-06-29
Fix incorrect current user name, closes #840

### LINKED ISSUE #840: [Bug report] Incorrect current user name
---------
I've  logged in as `a.kurakin` and, visually, I have `a.kurakin`'s permissions.

But I see:
image
So, I'm supposed to be `admin`.
(Note: `admin` isn't something special here. Superuserness doesn't matter here.)

Behavior is reproduced. Ready to debug.

screencast

(Note: I've logged in both as `admin` and `a.kurakin` before, using the same browser window.)

Your Environment
---------

*   Operating System: Ubuntu 18.04.4 LTS
*   Python Version Used: system Python is 3.6.9
*   When you install doccano: 95344e562908cbc87609f5e15c9bdf019f3c99b9 or b4c99b5503e16bf0477bf0325ef0ec2d571e3981 - the same
*   How did you install doccano (Heroku button etc): Docker Compose

Hironsan:


kuraga:
Confirm: patch works! Thanks!

Issue #658, opened 2020-04-01
[Enhancement request] Allow to import unlabeled datasets


kuraga:
Feature description
---------
Enhancement request: allow to import unlabeled datasets (update: via CSV).

Mentioned in #516, then merged to #521.

hmsaleem:
JSON was the only way I could import unlabeled datasets in the last version.

kuraga:
Update: via CSV

PR #861, opened 2020-06-30
Feature/import unlabeled dataset

cgill95:
Solving Issue #658 with this change.

Now it is possible to upload a CSV file that does not contain label for all text fields. Thus making it possible to upload partially labeled text together with labeled text.


app/api/utils.py
```diff
@@ -385,10 +385,10 @@ def parse_excel_csv_reader(reader):
                 yield data
                 data = []
             # Only text column
-            if len(row) == len(columns) and len(row) == 1:
-                data.append({'text': row[0]})
+            if len(row) <= len(columns) and len(row) == 1:
+                data.append({'text': row[0] })
```
Hironsan, line 389: Please remove an unnecessary space after `row[0]`.

app/api/utils.py
```diff
@@ -385,10 +385,10 @@ def parse_excel_csv_reader(reader):
...
             # Text, labels and metadata columns
-            elif len(row) == len(columns) and len(row) >= 2:
+            elif len(row) <= len(columns) and len(row) >= 2:
```
Hironsan, line 391: Simplify chained comparison:

```python
2 <= len(row) <= len(columns)
```

app/api/tests/data/example.csv
```diff
@@ -1,4 +1,4 @@
-text,label
-AAA,Positive
-BBB,Positive
+text,label,meta
+AAA
+BBB,Positive,The following is meta data
```
Hironsan, line 3: I tested the case of `text` and `meta` but it failed.

```
text,label,meta
text1,,meta1
text2,,meta2
```

cgill95: This now works and I added this specific to the testcase.

Issue #804, opened 2020-05-29
Execute refresh above the pagination-boundary example causes miss ordering


MikeVlad6:
How to reproduce the issue:
Scroll on any project with more than 10 documents (pagination limit) and hit shift+refresh. Upon reloading, instead of getting document ID (24 for example) will get the corresponding document on the first page, in this case, 4. 
This can be assessed with a print in store/documents.js, in the getDocumentList action, line 102, (console.log(payload)). 
I've been able to fix this by adding this.updateSearchOptions before this.getDocumentList in the BottomNavigator.vue. 

I don't know if this is of any help to you but it might so i'll leave it here.

icoxfog417:
What is the desirable specification you think? Please show us as-is and to-be.

MikeVlad6:
I've uploaded a demo here. This is made on the master branch, i just cloned and ran it. There are multiple components which are calling getDocumentList and the bug occurs when hitting refresh on the (11th or greater document) - it will bring you the corresponding document from 1 to 10 (if u hit refresh on the 11th it will bring you the first document, 12th - second document etc). You can reproduce the issue with any document sample greater than the pagination size (10).
I haven't been able to fix it so far so i can't provide a to-be specification and i've also modified alot on my fork so it might not be that relevant. 
I'll come back with a solution if or when i find one.

Thanks for the good job!
https://imgur.com/a/H7qjw3t

icoxfog417:
Thank you for providing the detail behavior. I understand this phenomenon as the following.

As-is

* When you execute refresh above the pagination-boundary example (11th or greater document), then the first example appears in front of the page. 

To-Be

* The example showed before the refresh and order of examples should be kept.

Or

* When executing refresh, back to top (1st) example.

icoxfog417:
@MikeVlad6 Could you please confirm the behavior of the latest doccano? This bug will be fixed.

PR #888, opened 2020-07-11
Fix wrong document list in text classification, closes #887

### LINKED ISSUE #887: Wrong document list in text classification with multiple projects
---------
If you work with two text classification projects at the same time and switch between them, the wrong document list is presented. 

doccano-wrong-docs4

Additionally, if you try to annotate the sentence, a wrong database record is inserted.

wrong-docs

Your Environment
---------
*   Using Docker with `docker-compose -f docker-compose.dev.yml up -d`
*   Original code from doccano/master on July 10, 2020, commit 690b688ded8775d079e44dee1f421ac4c36406fa

**I will submit a pull request the a proposed solution.**

lasp73:
Refers to #887 
Also helps #804 (refresh annotation page become more stable)

It seems that removing getDocumentList() from Paginator and BottomNavigator released this bug. 
So I decided to insert a getDocumentList() in TextClassification.vue but also defined **the offset**.

### DIFF
--- a/frontend/components/containers/annotation/EntityItemBox.vue
+++ b/frontend/components/containers/annotation/EntityItemBox.vue
@@ -11,7 +11,7 @@
 </template>
 
 <script>
-import { mapActions, mapGetters, mapState } from 'vuex'
+import { mapActions, mapGetters, mapMutations, mapState } from 'vuex'
 import EntityItemBox from '~/components/organisms/annotation/EntityItemBox'
 
 export default {
@@ -29,6 +29,12 @@ export default {
   },
 
   created() {
+    this.updateSearchOptions({
+      offset: this.offset
+    })
+    this.getDocumentList({
+      projectId: this.$route.params.id
+    })
     this.getLabelList({
       projectId: this.$route.params.id
     })
@@ -37,6 +43,7 @@ export default {
   methods: {
     ...mapActions('labels', ['getLabelList']),
     ...mapActions('documents', ['getDocumentList', 'deleteAnnotation', 'updateAnnotation', 'addAnnotation']),
+    ...mapMutations('documents', ['updateSearchOptions']),
     removeEntity(annotationId) {
       const payload = {
         annotationId,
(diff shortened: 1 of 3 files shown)


frontend/components/containers/annotation/Seq2seqContainer.vue
```diff
@@ -36,7 +36,17 @@ export default {
     }
   },
 
+  created() {
+    this.updateSearchOptions({
+      offset: this.offset
```
Hironsan, line 41: There is not an offset variable.

frontend/components/containers/annotation/EntityItemBox.vue
```diff
@@ -29,6 +29,12 @@ export default {
   },
 
   created() {
+    this.updateSearchOptions({
+      offset: this.offset
```
Hironsan, line 33: There is not an offset variable.

Hironsan:
Thanks for the PR.
But I reverted https://github.com/doccano/doccano/pull/868 temporary.

frontend/components/containers/annotation/Seq2seqContainer.vue
lasp73: @Hironsan, sorry, my mistake. Should I include "mapGetters('pagination', ['offset'])" in both?
I mean... do you think the change in PR is in the right way?

Hironsan: I think the bug(https://github.com/doccano/doccano/issues/887) is very important so I reverted https://github.com/doccano/doccano/pull/868. Therefore, https://github.com/doccano/doccano/issues/887 should have been resolved.

The below is a comment for "Should I include "mapGetters('pagination', ['offset'])" in both?"
- Yes. But the code is duplicated so I think the layouts/annotation.vue will be a better place.

br-dr3: I guess 868 is as important as this one. Maybe #868 should be applied and this bug should have a different approach, haven't it?

PR #964, opened 2020-08-27
Add timestamps to annotation serializers

### LINKED ISSUE #963: Expose timestamps associated with annotations with the API
The `..Annotation` models currently stores the `created_at` and `updated_at` fields but they are not exposed by the API.

I'd like to propose exposing them through the API so that downstream analysis can be conducted using simple API calls, like those made using [`doccano-client`](https://github.com/doccano/doccano-client), for instance.

setu4993:
Minor change that adds additional fields to all of the `..Annotation` models' serializers to expose timestamp information in API query results.

Fixes #963.

### DIFF
--- a/app/api/serializers.py
+++ b/app/api/serializers.py
@@ -177,7 +177,7 @@ class DocumentAnnotationSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = DocumentAnnotation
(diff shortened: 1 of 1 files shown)


setu4993:
@Hironsan @c-w @icoxfog417 : This PR is ready for review.

Issue #718, opened 2020-04-27
feature request: keyboard shortcut for approval button


mmisiewicz-yext:
Feature description
---------
Hello! I'd like to request a keyboard shortcut for the approval button. Ideally pressing this key would also advance to the next item in the queue. 

I could implement this myself if there's a place to do it!

Hironsan:
Solved in #980.

maxtrem:
@Hironsan 
This does not work anymore since the last update (at least for me). I found the "enter" solution to work great. 
Assuming this was removed by accident,  would it be possible to add this functionality again?

Thanks!

dpusat:
@Hironsan 
Hi - keyboard shortcut for approval isn't working for me as well. Has it been removed?

Thanks!

Hironsan:
Oops!

davidpomerenke:
I think it would be nice to have one key (perhaps enter) that marks the item as checked and at the same time moves to the next item.

(My use case is binary classification, which could be done with a single key press; but as I see it, I currently need 3 key presses in doccano: For selecting the label, for approving, and for moving to the next item. – Would be nice to cut that down to 2 key presses, or even one.)

Thansk for consideration!

hermlon:
I implemented (a probably pretty quick and dirty) solution for ImageClassification which automatically marks the image as accepted and opens the next page after assigning a label. Having to press 3 buttons (hotkey for label, enter to accept, right arrow for next) when only one is required is ridiculous when labeling about 1000 pictures. 

https://github.com/hermlon/doccano/commit/6d11d3cd3b9d70733c59eac9b2416448949fc049

PR #980, opened 2020-09-09
Adding keyboard shortcut for approval button

Collinbrown95:
Addressing issue #718

When user presses the ‘a’ key, the document toggles from ‘approved’ to ‘not approved’ or vice versa. The user is advanced to the next item in the queue once the key is pressed.

### DIFF
--- a/frontend/components/containers/annotation/ApproveButton.vue
+++ b/frontend/components/containers/annotation/ApproveButton.vue
@@ -2,7 +2,7 @@
   <v-tooltip bottom>
     <template v-slot:activator="{ on }">
       <v-btn
-        v-shortkey.once="['a']"
+        v-shortkey.once="['enter']"
         :disabled="disabled"
         class="text-capitalize ps-1 pe-1"
         min-width="36"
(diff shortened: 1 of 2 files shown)


frontend/components/containers/annotation/ApproveButton.vue
```diff
@@ -2,11 +2,13 @@
   <v-tooltip bottom>
     <template v-slot:activator="{ on }">
       <v-btn
+        v-shortkey.once="['a']"
```
Hironsan, line 5: The short key `a` is not appropriate if some label has the same short key.

Collinbrown95: Fair point. Do you have a preference on what a more appropriate short key would be?

Hironsan: I like the `enter` key.

The idea of using two keys is a good idea. But actually, I implemented the database scheme to be able to use a combination of two/three keys(ctrl, shift, ctrl+shift, and [0-9][a-z]) as a short key of a label. So we need to avoid using shortcut keys in that range.

Collinbrown95: Makes sense to me. I just committed a change to use ```enter``` instead of ```a```.

PR #1072, opened 2020-11-26
Deduplicate data on randomized order option

### LINKED ISSUE #933: Duplicate Documents seen after setting "Randomize Document Order Per User"
Hi,
While setting up a new project, I chose the option "Randomize Document Order Per User". Now when I import the dataset and every user started annotating, we are all seeing duplicate documents in the rows. This did not happen when we didnt select this setting. Even though duplicates are present the number of documents displayed on the page are same as the original no of documents(without duplicates). That means we are losing on some documents that are being replaced with these duplicate documents. 

Just to confirm we have checked and there are no duplicates in the file. So that means the issue is coming from Doccano and only when we are chosing the above mentioned setting.

We are currently using Doccano through deployment on Azure.

Any help is appreciated.
Thanks!

### LINKED ISSUE #1026: Duplication of the data occurs at the bottom of paging
---------
The data at bottom of every page is same, that also make reduplicate document when doing annotation.

image
image
image

Your Environment
---------
*   Operating System: ubuntu 16
*   Python Version Used: python3.6
*   When you install doccano: 2020-10-26
*   How did you install doccano (Heroku button etc): github

Hironsan:
fix #933 
fix #1026

### DIFF
--- a/app/api/views.py
+++ b/app/api/views.py
@@ -108,7 +108,6 @@ class StatisticsAPI(APIView):
             set_user_data[ind_obj['user__username']].add(ind_obj['document__id'])
         return {i: len(set_user_data[i]) for i in set_user_data}
 
-
     def progress(self, project):
         docs = project.documents
         annotation_class = project.get_annotation_class()
--- a/app/api/views.py
+++ b/app/api/views.py
@@ -1,5 +1,7 @@
 import collections
 import json
+import random
+
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.db import transaction
@@ -171,7 +173,9 @@ class DocumentList(generics.ListCreateAPIView):
 
         queryset = project.documents
         if project.randomize_document_order:
-            queryset = queryset.annotate(sort_id=F('id') % self.request.user.id).order_by('sort_id')
+            random.seed(self.request.user.id)
+            value = random.randrange(2, 20)
+            queryset = queryset.annotate(sort_id=F('id') % value).order_by('sort_id', 'id')
         else:
             queryset = queryset.order_by('id')
 


app/api/views.py
```diff
@@ -171,7 +172,9 @@ def get_queryset(self):
 
         queryset = project.documents
         if project.randomize_document_order:
-            queryset = queryset.annotate(sort_id=F('id') % self.request.user.id).order_by('sort_id')
+            random.seed(self.request.user.id)
+            value = random.randrange(2, 20)
+            queryset = queryset.annotate(sort_id=F('id') % value).order_by('sort_id', 'id')
```
kuraga, line 177: @Hironsan , really, how does it work?

Hironsan: Good question.

In fact, this is not a true randomization. I tried some randomization methods and it works well on a small dataset. However, if we use a large dataset, I found the randomization takes a long time.

That's why I decided to solve only the duplication problem. There is still an issue of true randomization.

kuraga: Ok, it's not truely random, but how do these lines work? I don't understand :)

And why does this patch work? We only replace `self.request.user.id` with `f(self.request.user.id)`, where `f` isn't random function. Yeah? `value = random.randrange(2, 20)` will be the the same for the same seed (`self.request.user.id`), woudn't it be?

kuraga: @Hironsan , Ok, it's not truely random, but how do these lines work? I don't understand :)

And why does this patch work? We only replace `self.request.user.id` with `f(self.request.user.id)`, where `f` isn't random function. Yeah? `value = random.randrange(2, 20)` will be the the same for the same seed (`self.request.user.id`), woudn't it be?

Issue #851, opened 2020-06-25
[Tiny enhancement request] Filename on JSONL export


kuraga:
Feature description
---------
doccano proposes to save dataset to `file.json` if you choose `JSONL` as export format.

Proposal:
1. Change `.json` extension to `.jsonl`.
2. Change `file` to more perceived filename (with project mention, as such as on labels export).

PR #1116, opened 2020-12-15
Enhancement/dynamic filename + changed json1 naming to jsonl

SwiftPredator:
Resolves: #851

### DIFF
--- a/frontend/store/projects.js
+++ b/frontend/store/projects.js
@@ -157,15 +157,18 @@ export const getters = {
   getExportFormat(state) {
     const csv = {
       type: 'csv',
-      text: 'CSV'
+      text: 'CSV',
+      suffix: 'csv'
     }
     const json = {
       type: 'json',
-      text: 'JSONL'
+      text: 'JSONL',
+      suffix: 'jsonl'
     }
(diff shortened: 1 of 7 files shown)


Hironsan (commented):
Thank you for your work.
It seems to me that this does not work for document classification and sequence to sequence project. It occurred 500 error.

frontend/store/projects.js
```diff
@@ -160,11 +160,11 @@ export const getters = {
       text: 'CSV'
     }
     const json = {
-      type: 'json',
+      type: 'jsonl',
```
fjen, line 163: This changes the path in `views.py:TextDownloadAPI()` and wrongly executes `JSONPainter.paint_labels()`, thus resulting in error 500 for the respective project types.

We need a better way to differentiate multiple "jsonl" formats or stick with "json" here.

fjen:
CSV export for document classification projects works fine.

SwiftPredator:
Thanks for your reviews. Should work now for all project types. In order to be able to set the suffix type independent of the format, I added a suffix variable in the project.js. @Hironsan @fjen

PR #1128, opened 2020-12-23
fix python3 compatibility problem

liangfu:
This is PR improved the python3 compatibility problem. Previously, it calls the python executable in the environment. This could be troublesome since many linux system has python2 as default executable in the environment, but the user might installed the python3 version of the package.

Changing 'python' to sys.executable would force the executable for django to be the same as doccano.

### DIFF
--- a/app/doccano/doccano.py
+++ b/app/doccano/doccano.py
@@ -1,5 +1,6 @@
 import argparse
 import os
+import sys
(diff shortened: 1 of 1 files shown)


liangfu:
cc @Hironsan please review.

Hironsan:
Thanks!

Issue #949, opened 2020-08-11
"Export only approved dataset" option


br-dr3:
Feature description
---------
We are trying to optimize jobs by having someone tagging and other approving documents and we gonna use it to train a nlp model and see how far it goes by the time..

It would be useful to have a filter in export dataset in order to have only the already approved documents.

PR #1138, opened 2021-01-04
Enhancement: Added option to export only approved datasets

SwiftPredator:
Resolves: #949


Hironsan:
@SwiftPredator Thanks. Merged.

Issue #706, opened 2020-04-20
Why is there a file size restriction to just 1 MB?


johann-petrak:
When I use docker-compose from the latest master version of doccano and when I try to import a JSONL file, I get the error message "File size should be less than 1 MB" and I cannot upload my file.
Why does this restriction exist?

I previously ran the docker version of doccano and there it was possible to import my file without problems. 

Is there a configuration option somewhere to disable this limitation?

kuraga:
Duplicate of #575, #514.

johann-petrak:
Thank you for those pointers to other issues!

#575 mentions a work-around by modifying some constant deep in the code which I think is not a solution and does not answer the question why such a small limit is the default. Also there is a report that the workaround did not fix the problem for the reporter.

#514 has been merged into #521 which seems to be about documenting the limitation but not telling why the upload size is so tiny. 

Ideally this should be configurable in a config file (as opposed to changing code) and the config option should be documented (unless there is an important reason why the upload has to be so small in which case I think it would not really be usable for some users).

kuraga:
> Ideally this should be configurable
Yes, I also think so.

I ask you to fill an issue via [Feature Request template](https://github.com/doccano/doccano/issues/new?template=04-request.md) and close this question. It will be more effectively.

johann-petrak:
OK, created #708

PR #1175, opened 2021-01-24
Increased max upload size to  100 MB.

ricjhill:
This is in response to  https://github.com/doccano/doccano/issues/706 which was closed for an unclear reason 

Feel free to reject the PR if it triggers some issue I  am not aware of.

Tested using docker-compose.prod.yml and docker-compose.dev.yml  with a  3MB dataset. If you want more testing before merging, then please tell me the filesize needed.

### DIFF
--- a/frontend/i18n/en/rules.js
+++ b/frontend/i18n/en/rules.js
@@ -29,7 +29,7 @@ export default {
   },
   uploadFileRules: {
     fileRequired: 'File is required',
-    fileLessThan1MB: 'File size should be less than 1 MB!'
+    fileLessThan1MB: 'File size should be less than 100 MB!'
   },
   passwordRules: {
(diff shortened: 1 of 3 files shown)


Hironsan:
Thanks! Merged.

xtfocus:
I still cannot upload labels json file in import labels of 1.2 MB 
image
Here's the code to create a pseudo label set of size 1.x MB:
```
import string
import random

fake_labels = [
    {
        "id": i,
        "text": ''.join(random.sample(string.ascii_letters, k = random.choice(range(8, 10))))
    } for i in range(23000)
]

with open("fake_labels.json", "w") as h:
    json.dump(fake_labels, h)
```

drmayu7:
as per version 1.8.4, i'm still having issue when importing label .json with size of 5mb

image

Issue #1257, opened 2021-03-18
export data as jsonl ,but the result is json.


13360806188:
in the datasets,I export data,when I choose jsonl and export ,the result is still json.I find the request url is 
v1/projects/2/docs/download?q=json&onlyApproved=false

not 
v1/projects/2/docs/download?q=jsonl&onlyApproved=false

abneetwats24:
Any fix for this?

abneetwats24:
Workaround 
using curl or postman, you can use following curl request for token 
```
curl --location --request POST 'http://172.16.10.6/v1/auth-token' \
--header 'Content-Type: application/json' \
--data-raw '{"username":"admin","password":"Admin123$"}'
```
then use this to get the jsonl response
```
curl --location --request GET 'http://172.16.10.6/v1/projects/1/docs/download?q=jsonl&onlyApproved=false' \
--header 'Authorization: Token <token from above api>' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json'
```
to download response in file use:
```
curl  -O data.jsonl --location --request GET 'http://172.16.10.6/v1/projects/1/docs/download?q=jsonl&onlyApproved=false' \
--header 'Authorization: Token <token from above api>' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json'
```

13360806188:
thanks.But it is a bug.I think it must be changed in the code .

abneetwats24:
Yes, it should be fixed.

MateusMarcuzzo:
thanks @abneetwats24, that may be a temporary solution for now.

I also need the fix

PR #1287, opened 2021-04-03
Fixed jsonl type error

SwiftPredator:
Many users reported that the wrong output occurs during exporting to the jsonl(textlabel) format. I investigated the error and think its because of a typo error in the format.ts file. 

See (#1267) for reference. Maybe also (#1257)


Hironsan:
Thanks.

PR #1301, opened 2021-04-13
Feature/tagsystem

### LINKED ISSUE #1044: Support tagging to distinguish multiple projects
Hi,

As title described. I think this feature will be helpful when:

- managing projects programmatically
- integrating doccano with other jobs/services. 

Here are my thoughts:

**UI:** 
Screen Shot 2020-11-13 at 5 00 04 PM

**DB Models:**
add one more table `api_tag`, with following columns:
```
id (PK)
text
created_at
project_id (FK)
text_color
background_color
```

**APIs:**
GET /projects/{project_id}/tags

response body may look like:
```
[
  {
    "id": 1,
    "text": "In Progress"
  },
  {
    "id": 2,
    "text": "Completed"
  }
]
```

POST /projects/{project_id}/tags
GET /projects/{project_id}/tags/{tag_id}
PUT /projects/{project_id}/tags/{tag_id}
PATCH /projects/{project_id}/tags/{tag_id}
DELETE /projects/{project_id}/tags/{tag_id}

## Usecases & Scenarios:
We want to integrate doccano with our interval jobs/services. As providing an automated platform, we want annotators explicitly mark a project as 'completed', so that certain downstream jobs can be triggered.

As we serving doccano for multiple internal teams, we also want to attach business-related information to the project, like model names, stakeholder teams, etc.

### LINKED ISSUE #1114: Mark project as done
Feature description
---------

Hi, is there a way for an annotator/project-admin to mark a project as done?
We could add a boolean field to the schema, and a checkbox in the front-end.

MrChrissie:
This feature enhances user experience through the addition of tags for projects.
Users can now create and delete tags for projects.
Resolves #1114 and partially implements the suggested functionality of #1044.

tags
tags1


app/api/migrations/0009_tag.py
```diff
@@ -0,0 +1,22 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Tag',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('text', models.TextField()),
+                ('project', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='tags', to='api.project')),
+            ],
```
Vydruth, line 20: The fields look good but I guess are not enough as per the request spec,the required fields are

> 1. id (PK)
> 2. text
> 3. created_at
> 4. project_id (FK)
> 5. text_color
> 6. background_color

as per the spec here: https://github.com/doccano/doccano/issues/1044

If you have another idea for text and background color please let me know. But I believe created_at/create_dt_tm is an important field to have.

Vydruth (changes requested):
Could you add some screenshots of how the UI looks with the tags? That would be helpful!

app/api/models.py
```diff
@@ -130,7 +130,6 @@ def get_storage(self, data):
         from .utils import Seq2seqStorage
         return Seq2seqStorage(data, self)
 
-
```
Vydruth, line 133: NIT: unnecessary line removal

frontend/plugins/services.ts
```diff
@@ -84,9 +88,11 @@ const plugin: Plugin = (context, inject) => {
   const seq2seq = new Seq2seqApplicationService(seq2seqRepository)
   const option = new OptionApplicationService(optionRepository)
   const config = new ConfigApplicationService(configRepository)
+  const tag = new TagApplicationService(tagRepository)
   const template = new TemplateApplicationService(templateRepository)
   const auth = new AuthApplicationService(authRepository)
-  
+
+
```
Vydruth, line 95: NIT: Looks like an unintentional spacing change.

frontend/components/project/ProjectList.vue
```diff
@@ -76,8 +83,9 @@ export default Vue.extend({
         { text: this.$t('generic.description'), value: 'description' },
         { text: this.$t('generic.type'), value: 'projectType' },
         { text: 'Updated', value: 'updatedAt' },
+        { text: 'Tags', value: 'tags'}
       ]
     }
-  } 
+  },
```
Vydruth, line 89: You don't need a "," here. :)

app/api/migrations/0009_tag.py
fjen: What would be the purpose of a creation date?

Vydruth: I believe it is a good idea to log when a resource is created or deleted in general. Today the purpose of the tags are limited but in the future, we could add a page(view) showing all projects for a specific tag, and on that page, we might want to sort the projects by most-to-least (or vice versa) recently tagged projects. This might be helpful for a user to figure out the priority based on the date last tagged. 

Do let me know your thoughts if you feel the above-mentioned are not valid use cases or are covered as part o other issues.

Hironsan:
Please select "Resolve Conversation" if the problem is solved.

Vydruth:
@Hironsan I don't see a "resolve conversation" button, I am just waiting on a response on why the `created_at` is not added as mentioned per spec #1044? 

I am not very worried about the label colors, although I think it would be good to have. I'd like to know if there was a discussion that happened somewhere about not needing these fields? If not, maybe it's a point of discussion [here](https://github.com/doccano/doccano/pull/1301#discussion_r612885803).

app/api/migrations/0009_tag.py
MrChrissie: The additional of a creation date is not currently needed and will be of limited use in the future.
The priority of a project can also be indicated by tags instead of your most recently tagged project criterion.
After all one might expand the tag functionality by color or created_at in the future if needed.

Vydruth: Well, I agree that it is of limited use and you do have a point about priority being represented as a tag, did not think of it that way. Sounds good to me!

Vydruth (approved):

Vydruth:
@Hironsan: this branch can be merged! :)

Hironsan:
@Vydruth Merged. Thanks!

Issue #1238, opened 2021-03-05
[Enhancement request] Using without Internet


kuraga:
Feature description
---------
Haven't an ability to set doccano in a closed environment (i.e. without Internet): doccano uses on-line Google Fonts (UPD: and some other issues).
Feature request: provide using without internet.

kuraga:
My research by `git grep http` on v1.2.1:
```
frontend/nuxt.config.js:      { src: 'https://use.fontawesome.com/releases/v5.0.6/js/all.js' }
frontend/nuxt.config.js:          'https://fonts.googleapis.com/css?family=Roboto:300,400,500,700|Material+Icons'
app/server/templates/admin.html:<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/styles/vs2015.min.css">
app/server/templates/base.html:  <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.0.13/css/all.css" integrity="sha384-DNOHZ68U8hZfKXOrtjWvjxusGo9WQnrNx2sqG0tfsghAvtVlRW3tvkXWZh58N9jp"
app/server/templates/base.html:  <link href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,700" rel="stylesheet">
app/server/templates/base.html:  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bulma/0.7.1/css/bulma.min.css" crossorigin="anonymous">
app/server/templates/base.html:  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bulma-extensions@4.0.1/bulma-divider/dist/css/bulma-divider.min.css" crossorigin="anonymous">
app/server/templates/base.html:  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bulma-extensions@4.0.1/bulma-checkradio/dist/css/bulma-checkradio.min.css" crossorigin="anonymous">
app/server/templates/base.html:  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bulma-extensions@4.0.1/bulma-tooltip/dist/css/bulma-tooltip.min.css" crossorigin="anonymous">
app/server/templates/index.html:<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/Swiper/4.3.3/css/swiper.min.css">
[...]

PR #1322, opened 2021-04-21
Add offline deployment support

j-frei:
This pull request adds a small set of scripts to deploy Doccano within an offline network environment.

This work was done by myself for internal use but I think it could be helpful for other users that are considering Doccano as an annotation tool in offline environments.

The scripts mainly do the following tasks:
 - For preparation: Build the Docker images
 - For preparation: Load required APT packages
 - For preparation: Download referenced JS/CSS/fonts from CDNs and provide them as static assets
 - For offline deployment: Install APT packages and load Docker images

The focus of this pull request is to keep the other repository code as decoupled as possible. Due to this approach, there is no need for maintaining a hardcoded list of used CDN resources.

Your feedback would be highly appreciated. :)


Hironsan (commented):
I want you to do two things:

- Please move script files to `/tools/offline_deployment` directory.
- Please move README.md to `docs/advanced` directory.

j-frei:
Should be fine now. I ran the updated scripts successfully on my vm test instances.

Hironsan:
Thanks!

close #1238

PR #1323, opened 2021-04-22
Changed "doccano webserver -p" argument to "--port"

lucasrdrgs:
See diffs. This md file goes through the setup process using pip. However, it says to use `doccano webserver -p 8000`, but the parameter -p is not valid. You must use --port instead. That's what I have changed.


Hironsan:
Thanks

PR #1340, opened 2021-04-30
fix download dataset problems

pq4362050:
download dataset to a csv file, but I get an empty file in zip
download json file or jsonL file contains chinese, it shows \uxxxx but not chinese


backend/api/views/download/writer.py
```diff
@@ -135,7 +136,7 @@ def create_line(self, record):
             'data': record.data,
             'label': record.label,
             **record.metadata
-        })
+        },ensure_ascii=False)
```
Hironsan, line 139: missing whitespace after ','. Please add whitespace.

Issue #1307, opened 2021-04-19
Can't create new Custom REST Request model


gika815:
Trying to create new Auto-labling model (Custom REST request). 
When I press the button "Test" I have "NameError: Custom REST Request is not found" error in Docker logs

Forbidden: /v1/me

[19/Apr/2021 07:55:24] "GET /v1/me HTTP/1.1" 403 58

Forbidden: /v1/me

[19/Apr/2021 07:55:25] "GET /v1/me HTTP/1.1" 403 58

[19/Apr/2021 07:55:34] "POST /v1/auth/login/ HTTP/1.1" 200 50

[19/Apr/2021 07:55:34] "GET /v1/me HTTP/1.1" 200 47

[19/Apr/2021 07:55:34] "GET /v1/projects HTTP/1.1" 200 400

[19/Apr/2021 07:55:39] "GET /v1/projects/1 HTTP/1.1" 200 357

[19/Apr/2021 07:55:40] "GET /v1/projects/1 HTTP/1.1" 200 357

[19/Apr/2021 07:55:41] "GET /v1/projects/1 HTTP/1.1" 200 357

[19/Apr/2021 07:55:41] "GET /v1/projects/1/auto-labeling-configs HTTP/1.1" 200 2

[19/Apr/2021 07:55:42] "GET /v1/projects/1/auto-labeling-templates HTTP/1.1" 200 127

[19/Apr/2021 07:55:42] "GET /v1/projects/1/labels HTTP/1.1" 200 342

[19/Apr/2021 07:55:44] "GET /v1/projects/1/auto-labeling-templates/GCP%20Entity%20Analysis HTTP/1.1" 200 1491

[19/Apr/2021 07:55:45] "GET /v1/projects/1/auto-labeling-templates/Custom%20REST%20Request HTTP/1.1" 200 531

Internal Server Error: /v1/auto-labeling-parameter-testing

Traceback (most recent call last):

File "/src/app/api/views/auto_labeling.py", line 113, in post

model = RequestModelFactory.create(model_name, model_attrs)

File "/root/.local/share/virtualenvs/src-NVTF7jWz/lib/python3.8/site-packages/auto_labeling_pipeline/models.py", line 21, in create

subclass = cls.find(model_name)

File "/root/.local/share/virtualenvs/src-NVTF7jWz/lib/python3.8/site-packages/auto_labeling_pipeline/models.py", line 30, in find

raise NameError(f'{model_name} is not found.')

[...]

PR #1368, opened 2021-05-14
Remove unneeded builder call

rolisz:
This a PR to fix part of the issues from #1307, which caused failures when trying to create Custom REST Request models.


Hironsan:
Thanks!

PR #1384, opened 2021-05-28
Relations annotation added to SequenceLabeling task

DavideAllavena:
With this PR we introduce a new feature: the relation between annotations.

Each annotation can be linked to another with a _"relation"_ (or link) which has a direction and a type (types can be added/modified in the "relations" page). Only one relation can exists from the same pair of source and target.

The relations are drawn as colored arrows.

# How to use it
To try the new feature you should, first, create some relation type in the "relations" page. The relation semantics is free.
Then, to create a relation you can click the R icon (a new icon at the top right of any annotation).

This will open a menu showing the list of the outgoing relations (if any) and the option "new relation...".
Here the user can delete any existing relation by clicking the trashcan icon.

If the "new relation..." is selected the menu will ask to choose the relation type.
After that the R icon from the source annotation is shown in a blue background, and the other annotation, at the mouse hover, will show a + (plus) icon. Just click any + icon to select the relation target. This will create the relation and an arrow will appear at screen.

To abort the relation operation the user can click the R button again whenever he/she wants.


Hironsan:
Thanks. Please pass CI.

DavideAllavena:
> Thanks. Please pass CI.

Done some modification, I need CI approval run tests on dockerhub.

DavideAllavena:
All test are green! To do the dockerhub part i needed to change line 25 in `.github/workflows/publish-image.yml`

Hironsan:
Thanks! Merged.

sorenmc:
In my team we have tried this feature, but we can't seem to find the relation in the exported data.
How does this work?

DavideAllavena:
The data are stored in the database but still not included in the export. I think the maintainers should study a clever format to add this data to the export.

sorenmc:
Thanks for the quick reply!

PR #1432, opened 2021-07-02
Fixing Data Annotation Issues

zanussbaum:
When uploading datasets, the code uses a `bulk_create` to upload
Examples and Labels. It then filters the data from the database
based on when it was created. However, [Django doesn't enforce
the list order when calling filter](https://stackoverflow.com/questions/7163640/what-is-the-default-order-of-a-list-returned-from-a-django-filter-call) unless ordering is specified. The previous behavior mismatched labels
and examples. When this was shown in the UI, the data would show labels
for incorrect examples (i.e. a label for message #2 would be shown on
message #1). This fix enforces that the data is returned in the order
it was inserted so that the data, label pair is as expected. This was slightly hard to detect as the ordering returned from Postgres/Django was variable and would sometimes return the expected results but other times not.

### DIFF
--- a/backend/api/models.py
+++ b/backend/api/models.py
@@ -1,6 +1,8 @@
 import string
 from typing import Literal
 
+from django import db
+
 from auto_labeling_pipeline.models import RequestModelFactory
 from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
@@ -92,7 +94,7 @@ class ImageClassificationProject(Project):
 
 
 class Label(models.Model):
-    text = models.CharField(max_length=100)
+    text = models.CharField(max_length=100, db_index=True)
     prefix_key = models.CharField(
         max_length=10,
         blank=True,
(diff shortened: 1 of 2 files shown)


rolisz:
I think I just ran into this issue as well. Is the data in the database corrupted or if I apply your fix it will work again?

zanussbaum:
I believe I just created a fresh database/Django server. I believe adding the ordering changes the schema slightly, but there may be a way with `python manage.py makemigrations` or something like that so you don't have to create a new DB. 

I don't have the greatest Django experience but from what I could tell in the logs, it seems that it's possible to update the database without deleting everything and starting from scratch. 

I don't know about the ordering **IN** the database however. From what I could tell when I was developing locally, the files I uploaded were saved in the same way I had uploaded. However, I'm not sure of how they looked in the database. It may be worth querying the DB directly to see.

PR #1442, opened 2021-07-08
Remove unneeded query

rolisz:
Bulk create returns the created objects in the same order as they have been added.

In Postgres, the query was wrong, because ordering was not guaranteed and sometimes it returned the ids in the wrong order than they were inserted.

This is a simpler fix for #1432, without requiring changes to the model.

### DIFF
--- a/backend/api/tasks.py
+++ b/backend/api/tasks.py
@@ -62,10 +62,8 @@ class DataFactory:
             self.data_class(project=project, **example.data)
             for example in examples
         ]
-        now = datetime.datetime.now()
-        self.data_class.objects.bulk_create(dataset)
-        ids = self.data_class.objects.filter(created_at__gte=now)
-        return list(ids)
+        results = self.data_class.objects.bulk_create(dataset)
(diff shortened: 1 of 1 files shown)


zanussbaum:
Does this work also when exporting the data? I know that there's a [filter statement](https://github.com/doccano/doccano/blob/cc364068485264cacbc5516cc7e4b4b5b6b7b98a/backend/api/views/download/repositories.py#L78) to get the data, but we filter based on `project_id`. Curious if this issue would still occur.

rolisz:
It should work. Attaching the labels to the wrong documents happened at import time. At export time each document is exported with its own labels, as they are in the database.This pull request makes sure that the documents and labels are aligned at import time, so that they are saved properly in the database.On Jul 10, 2021 2:48 AM, Zach Nussbaum ***@***.***> wrote:
Does this work also when exporting the data? I know that there's a filter statement to get the data, but we filter based on project_id. Curious if this issue would still occur.

Issue #935, opened 2020-07-29
New user type: temporary user


Hironsan:
I'm thinking of dividing users into the following four types:

- Superuser
- Staff
- General user
- (New) Temporary user

Description:
- Superuser is the administrator of the Django application, not the project administrator of doccano. Basically it is created by Django's `createsuperuser` command.
- Staff can sign in to the admin site just like a superuser, but with more limited permissions. A superuser or a staff can create them from the admin site.
- General users cannot sign in to the admin site, but they can create projects. A superuser or a staff can create them from the admin site.
- Temporary user is a user who can neither sign in to the admin site nor create a project. This user is mainly for annotators(cloud workers). The project administrators of doccano can create them from the project members page.

The permissions per user types are as follows:

  | Super User | Staff | General User | Temporary User
-- | -- | -- | -- | --
Admin Site |   |   |   |  
　Login | ✓ | ✓ |   |  
　Other operations | ✓ | dependes on permissions |   |  
ProjectList |   |   |   |  
　Create | ✓ | ✓ | ✓ |  
　List | ✓ | ✓ | ✓ | ✓

This will help resolve Issue #920. In addition, this will allow staff and general users to create projects.

P.S.

- Within each project, the roles assigned to each user determine the permissions.
- There are three roles: Administrator, Annotator, and Auditor. The auditor is the current Approver.

  | Administrator | Annotator | Auditor
-- | -- | -- | --
ProjectDetail |   |   |  
　View | ✓ | ✓ | ✓
　Update | ✓ |   |  
　Delete | ✓ |   |  
LabelList |   |   |  
　Create | ✓ |   |  
　List | ✓ | ✓ | ✓
　Import | ✓ |   |  
LabelDetail |   |   |  
　View | ✓ | ✓ | ✓
　Update | ✓ |   |  
　Delete | ✓ |   |  
DocumentList |   |   |  
　Create | ✓ |   |  
　List | ✓ | ✓ | ✓
　Import | ✓ |   |  
　Export | ✓ |   |  
DocumentDetail |   |   |  
　View | ✓ | ✓ | ✓
　Update | ✓ |   |  
　Delete | ✓ |   |  
AnnotationList |   |   |  
　Create | ✓ | ✓ | ✓
　List | ✓ | ✓ | ✓
AnnotationDetail |   |   |  
　View | ✓ | ✓ | ✓
　Update | ✓ | ✓ | ✓
　Delete | ✓ | ✓ | ✓
UserList |   |   |  
[...]

PR #1453, opened 2021-07-20
Restrict project creation to staff users

youichiro:
## Issue
https://github.com/doccano/doccano/issues/935

## What this PR do
- restrict project creation to staff users.
- don't display create and delete project buttons if login user is not a staff.

## Why
- Annotator and approver don't need to create a project.
  - If necessary, a superuser changes the general user to staff.
- Issue #935 says to add a new user type, but I implement without changing DB schema for now.
  - Staff can login to the admin page, but they can't do anything unless a superuser grants them some permissions.
  - If you do not want to change a general user to staff, we will need to add a new (guest) user type and change the restrictions.

## Screenshot
Create and delete buttons are hidden when login user is not a staff.
スクリーンショット 2021-07-21 1 14 46

## Concern
If you already let general users creating projects, you will need to update that users information.


frontend/store/auth.js
```diff
@@ -46,13 +53,16 @@ export const actions = {
       commit('setAuthenticated', true)
       commit('setUsername', user.username)
       commit('setUserId', user.id)
+      commit('setIsStaff', false)
```
Hironsan, line 56: Is `user.isStaff` correct?
Even if the user is staff, the create button is hidden:

backend/api/views/project.py
```diff
@@ -17,7 +18,10 @@ def get_queryset(self):
         return self.request.user.projects
 
     def perform_create(self, serializer):
-        serializer.save(users=[self.request.user])
+        if self.request.user.is_staff:
+            serializer.save(users=[self.request.user])
+        else:
+            raise ProjectCreationPermissionDenied()
```
Hironsan, line 24: Maybe non-staff user but project admin can delete projects.
Defining staff permission and setting it to `permission_classes` may be a safer way.

frontend/store/auth.js
youichiro: fixed. https://github.com/doccano/doccano/pull/1453/commits/a6397eb5ec0071e544fd470a0eb778d17d8bfb3e

backend/api/views/project.py
youichiro: Thanks for your comments.
Changed to use `permission_classes`. https://github.com/doccano/doccano/pull/1453/commits/ed37696a67a0e1dc90fa5fdcd0c566e223a7b6c0

Hironsan:
Merged. Thanks!

PR #1505, opened 2021-09-27
fix Username/Password bug in aws cloudformation template

whn09:
Since code update, the username and password part in docker-compose.prod.yml have been changed. I fixed Username/Password bug in aws cloudformation template.


whn09:
Not best practice

Issue #544, opened 2020-01-20
Feature Request: Support Nested Named Entity Recognition


alokpadhi:
Feature description
---------

Currently NER tagging only provides to label single entity at a time. Is it possible to do entity inside entity (nested entity).
For example inside an entity personal info, an entity name can be placed.
My name is xxx and I live in yyy. Here the whole sentence is personal info but the xxx is a name entity.

icoxfog417:
There are some variations of entity overwrapping. We have to deal with these cases carefully to implement the annotation feature.

Reference: [Recognizing Complex Entity Mentions: A Review and Future Directions](https://www.aclweb.org/anthology/P18-3006/)

Hironsan:
Hello @alokpadhi ,

Do you know any annotation tool which can annotate nested NER?
If you know it, please tell me for my reference.

alokpadhi:
> Hello @alokpadhi ,
> 
> Do you know any annotation tool which can annotate nested NER?
> If you know it, please tell me for my reference.

For that matter you can use [Inception](https://inception-project.github.io/), [Dataturks](https://dataturks.com/), [brat](http://brat.nlplab.org/)

I found inception and dataturks user friendly. But I will be very happy if Doccano comes with this feature asap.

scheiblr:
> Hello @alokpadhi ,
> 
> Do you know any annotation tool which can annotate nested NER?
> If you know it, please tell me for my reference.

[Anafora](https://github.com/weitechen/anafora) could be interesting, too.

alokpadhi:
> > Hello @alokpadhi ,
> > Do you know any annotation tool which can annotate nested NER?
> > If you know it, please tell me for my reference.
> 
> [Anafora](https://github.com/weitechen/anafora) could be interesting, too.

I see, it's quite an interesting tool. and It looks more advance tool for labeling schema than the ones I have mentioned earlier.

MikeVlad6:
I needed this feature for myself so i made an implementation which is fine for my use. Unfortunately i had to remove the update functionality because the outer NER would always juxtapose the inner NER in terms of v-menu. 
I've also added the functionality to "round" selection to whole words.

Hironsan:
Rough design idea:

![image](https://user-images.githubusercontent.com/6737785/88457207-885b9480-cebf-11ea-92fe-d9608e6f1610.png)

Hironsan:
Coming soon...
[...]

PR #1511, opened 2021-10-08
Support nested named entity recognition, relation labeling, RTL language, improve speed and grapheme mode

Hironsan:
## Nested mode

Of course, you are able to annotate text without using nested mode.

## RTL Language

In the case of mixed RTL and LTR Language:

## Grapheme mode

In JavaScript, there is not always a one-to-one relationship between string characters and what a user would call a separate visual "letter". Some symbols are represented by several characters. This can cause issues when you annotate the part of the text in sequence labeling.

For example, emoji characters like "🌷","🎁","💩","😜" and "👍" are represented by two JavaScript characters each (high surrogate and low surrogate). That is,

```javascript
"🌷".length == 2
```

The combined emoji is even longer:

```javascript
"🏳️‍🌈".length == 6
```

Therefore, If your text includes such characters, the start/end offset indicates codepoint offset which is different from a "visual offset".

So, I implemented the "Grapheme mode" which counts the character per grapheme cluster. As an example, let's say you select the text as follows:

image

With the grapheme mode, the offset is `(1, 3)`. On the other hand, without the mode, the offset is `(4, 12)` because it counts the codepoints.

## Large Document

I introduced the virtual rendering system to handle a large document. Now, we are able to annotate it(I tested about 1,000,000 characters document).

### DIFF
--- a/frontend/components/project/FormCreate.vue
+++ b/frontend/components/project/FormCreate.vue
@@ -80,7 +80,7 @@
                     grapheme clusters
                   </a>
                 </template>
-                Like emoji(🌷, 💩, and 👍)
+                Like emoji(🌷, 💩, and 👍), CRLF(\r\n), and so on.
               </v-tooltip>
               as one character
             </div>


Hironsan:
closes #544

kris927b:
> Hi, I use this 1.5.0 version, the relation labeling is not working in sequence labeling. And also creation of relation is not working. Could anyone check the issue?

I have the same issue when using the production docker-compose version. What could be causing this issue?

Hironsan:
I have a plan to extract relation labeling as an independent task. So this PR doesn't include a relation labeling feature.

Status: I'd like to implement that feature, but I haven't gotten around to it because I'm fixing other bugs.

zhouao0314:
Hello @Hironsan , we are trying to use this version, but we need to use the relationship tag function. When will this function be merged?

zhouao0314:
Hello, @Hironsan, we are trying to use this version, but we need to use the relationship tag function. When will this function be merged?

anyi3c:
Hi, @Hironsan, we are trying to use this version, but we need to use the relationship Labeling function. When will this function be merged?

Srikanthv466:
I would like to annotated same entity with multiple labels, so how do we setup nested mode?

PeterPanUnderhill:
Is there an approximate time we can expect? thx

PeterPanUnderhill:
> Is there an approximate time we can expect? thx

@Hironsan

fernandrez-ai:
@Hironsan how can I use this? Can't find any example on how to use nested mode

Hironsan:
Please select "Allow overlapping spans" on creating a sequence labeling project.

Issue #1398, opened 2021-06-03
Validate Example by typing "Enter"


maxtrem:
Previously it was possible to validate or "Check" an annotation example by typing the Enter key. This feature was extremely useful for validation, but it disappeared with one of the recent updates.

Would it be possible to add this feature again?

Thanks!

PR #1562, opened 2021-11-17
Add keyboard shortcut back to accept button

FynnSu:
It seems like the "enter" key keyboard shortcut for the accept/review button functionality was lost at some point. See discussion in  #718 and issue #1398. This pull request restores the functionality


Hironsan:
Thanks!

PR #1572, opened 2021-11-23
persian language added

hiddenespy:
persian language added


Hironsan:
Thanks! But I think you forgot to commit under `frontend/i18n`.

PR #1573, opened 2021-11-23
Patch 1

hiddenespy:
persian language added


Hironsan:
Temporarily closed. Because there is no frontend/i18n/fa.

PR #1712, opened 2022-03-02
Fix state of label selection on LabelingMenu for sequenceLabeling

jhoobergs:
Currently labeling two entities with the same label by choosing it from the
autocomplete, didn't work. Labeling the first entity worked, but when you
tried to label a new entity with the same label chosen from the
autocomplete, nothing happened.


Hironsan:
Thanks! Really helpful.

Issue #1595, opened 2021-12-13
psycopg2.errors.UniqueViolation: could not create unique index "api_example_uuid_key"


zzOzz:
cf:
https://ajrbyers.medium.com/django-migrations-and-uuid-fields-4e3b27701e4a

Making staticfiles
Initializing database
Operations to perform:
  Apply all migrations: admin, api, auth, authtoken, contenttypes, django_celery_results, django_drf_filepond, sessions, social_django
Running migrations:
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UniqueViolation: could not create unique index "api_example_uuid_key"
DETAIL:  Key (uuid)=(6ca65fef-955e-4a98-8088-d2c0342b1474) is duplicated.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "manage.py", line 15, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 89, in wrapped
    res = handle_func(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/commands/migrate.py", line 244, in handle
    post_migrate_state = executor.migrate(
  File "/usr/local/lib/python3.8/site-packages/django/db/migrations/executor.py", line 117, in migrate
    state = self._migrate_all_forwards(state, plan, full_plan, fake=fake, fake_initial=fake_initial)
  File "/usr/local/lib/python3.8/site-packages/django/db/migrations/executor.py", line 147, in _migrate_all_forwards
    state = self.apply_migration(state, migration, fake=fake, fake_initial=fake_initial)
[...]

PR #1716, opened 2022-03-03
Fix migration when adding uuids

rolisz:
This is a fix for #1595.

What happened was that if there was data in the DB, a single uuid4 value would be created, so the migration would fail because of the unique constraint. 

The change is to first create the field, without a unique constraint and with null values, run a Python function to create uuid4s for each Example, and then change the field to have the unique constraints.

### DIFF
--- a/backend/api/migrations/0017_example_uuid.py
+++ b/backend/api/migrations/0017_example_uuid.py
@@ -5,6 +5,13 @@ import uuid
 from django.db import migrations, models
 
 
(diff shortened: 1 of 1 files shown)


Hironsan:
Great! Thanks.

I fixed the code a bit according to the following:

- [Migrations that add unique fields](https://docs.djangoproject.com/en/dev/howto/writing-migrations/#migrations-that-add-unique-fields)

PR #1862, opened 2022-06-10
Quick add relationship without mode-switch

leonardlin:
Problem: SequenceLabelling with Relationship requires explicit switch to "relation" mode with mouse
In order to annotate relationships you have switch to "relation" Mode with the mouse.
This slows down the labeling process.

Solution:
This PR allows to click two Entities and create a relationship without explicitly switching to "relation" mode.
If a relationship-type is preselected, the annotation process is even faster and the relationship type is automatically applied.
If no relationship-type is preselected, the relationship-label menu appears.


leonardlin:
I fixed the prettier warnings.
Ready to rerun. thanks

PR #1869, opened 2022-06-22
fix de, fr and zh messages

morioka:
regarding #1175, max upload size was increased  from 1MB to 100 MB.

But, only English message was fixed.  

```
de/rules.js:    fileLessThan1MB: 'Dateigröße muss kleiner als 1 MB sein!'
en/rules.js:    fileLessThan1MB: 'File size should be less than 100 MB!'
fr/rules.js:    fileLessThan1MB: 'La taille du fichier doit être inférieure à 1MB'
zh/rules.js:    fileLessThan1MB: '文件大小必须小于 1 MB!'
```

### DIFF
--- a/frontend/i18n/de/rules.js
+++ b/frontend/i18n/de/rules.js
@@ -29,7 +29,7 @@ export default {
   },
   uploadFileRules: {
     fileRequired: 'Datei(en) werden benötigt',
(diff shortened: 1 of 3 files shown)


Hironsan:
Thanks!

PR #1927, opened 2022-07-31
fix: install guide not acessible in the documentation website.

abdelhak2406:
In the developer guide section at the end of the backend the link that
redirects to the `installation guide` was displaying : 404 - Not found


Hironsan:
Thanks!

PR #1934, opened 2022-08-04
lang: better translation for french

abdelhak2406:
* improve current french translations
* add new translations for other words


Hironsan:
Thanks! Merged.

PR #2008, opened 2022-09-19
Add option to serve with URL prefix

### LINKED ISSUE #1878: Deploy doccano in a subpath
I know this is an old issue. But after some attempts I still failed to deploy doccano under some subpath. 

And I know there is another label tool named [Label Studio](https://labelstud.io/), it supports subpath by command like this:

    $ label-studio --host https://example.com/by/subpath/of/label-studio

So I just read its code and try to change the `FORCE_SCRIPT_NAME` in doccano but seems not working. 

https://github.com/heartexlabs/label-studio/blob/70f12eb96860709bb53c10bf1aa9e1257adef166/label_studio/core/settings/base.py#L45

Here is some attemp I did: 

https://github.com/aisensiy/doccano/commit/66b87e40e742f72dc7f494c145bb859490d4047b

I really want to make doccano support the subpath. Hope some one give some tips about this.

## My Environment

- Operating System: Mac OS 12.4 M1
- Python Version Used: 3.8
- When you install doccano: June 24
- How did you install doccano (Heroku button etc): `pip install doccano`

## Related issues:

- https://github.com/doccano/doccano/issues/1129
- https://github.com/doccano/doccano/issues/380
- https://github.com/doccano/doccano/issues/967

Vinno97:
Added the option to run Doccano with a URL prefix instead of always running from the domain root. The prefix is configured by specifying a `BASE_URL` environment variable. This PR fixes #1878

To run Doccano under `/doccano/`, set `BASE_URL` to `/doccano/`:

```sh
export BASE_URL=/doccano/
```

When we now compile the front-end, all routes are prefixed with BASE_URL.

We also need to let Django know about our new prefix. This is done using the same BASE_URL environment variable but needs to be set during runtime. I could not find a clean method to configure them both during runtime.

Since such a configuration will most likely only be used in combination with a load balancer, I also added the option to set `USE_X_FORWARDED_HOST` using environment variables.

Note that I did not write any tests for this (yet), both because I am unsure how this should be done and because I would first like to know if this PR would be acceptable.

### DIFF
--- a/frontend/components/configAutoLabeling/form/FileField.vue
+++ b/frontend/components/configAutoLabeling/form/FileField.vue
@@ -35,7 +35,7 @@ export default {
     return {
       myFiles: [],
       server: {
-        url: '/v1/fp',
+        url: `${process.env.baseUrl}/fp`,
         headers: {
           'X-CSRFToken': Cookies.get('csrftoken')
         },
--- a/frontend/pages/projects/_id/dataset/import.vue
+++ b/frontend/pages/projects/_id/dataset/import.vue
@@ -119,7 +119,7 @@ export default {
       ],
       requiredRules: [(v) => !!v || 'Field value is required'],
       server: {
-        url: '/v1/fp',
+        url: `${process.env.baseUrl}/fp`,
         headers: {
           'X-CSRFToken': Cookies.get('csrftoken')
         },


Hironsan:
Thanks!

Please write some test cases. After that, I'll merge this PR.

Vinno97:
Great, will do. Any tips about where to start for someone unfamiliar with the project (and Django in general)? I expect I should extend `backend/api/tests/test_config.py`?

lyingcrow:
Still in progress?

Vinno97:
This hasn't been on my radar for quite a long time. The project I needed this fix for, was finished up a while ago. I'd be willing to write some tests for it, but I still have no clue where to start.

backend/config/settings/base.py
```diff
@@ -245,7 +249,7 @@
 
 # User media files
 MEDIA_ROOT = env("MEDIA_ROOT", path.join(BASE_DIR, "media"))
-MEDIA_URL = "/media/"
+MEDIA_URL = BASE_URL + "/media/"
```
Hironsan, line 252: `BASE_URL + "media/"` ?

backend/config/urls.py
```diff
@@ -63,3 +63,6 @@
     path("swagger/", schema_view.with_ui("swagger", cache_timeout=0), name="schema-swagger-ui"),
     re_path("", TemplateView.as_view(template_name="index.html")),
 ]
+
+if settings.BASE_URL:
+    urlpatterns = [path(f'{settings.BASE_URL.strip("/")}/', include(urlpatterns))]
```
Hironsan, line 68: These lines seem to make the URL (e.g. <http://127.0.0.1:8000/v1/projects>) inaccessible. `settings.BASE_URL` is evaluated as `True`.

PR #2031, opened 2022-10-26
Add user creation API

Hironsan:
In this PR, I will add a user creation endpoint(`/v1/users/creation`). We can use them to add users programmatically. The payload is as follows:

```bash
{
    "username": str,
    "password1": str,
    "password2": str
}
```

The `password2` is the confirmation purpose. So if we add the user creation method to [doccano-client](https://github.com/doccano/doccano-client), we only give `username` and `password` to it:

```python
client.create_user(username="", password="")
```


david-engelmann:
@Hironsan This looks perfect, once merged, create_user can be added to the client and documented

Issue #1883, opened 2022-06-30
Include Example number in the Project Comments view


ljades:
Feature description
---------
On the backend API and in the dataset annotation interface, the Comments are associated with specific Examples. But in the Project Comments view, the Example association is unclear--all the comments are grouped together.

Can the Project Comments view tab be improved to detail Examples, maybe even sort or group by Example?

Thanks!

ljades:
Using the Docker Compose setup, running on Mac OS X Catalina 10.15.7

PR #2152, opened 2023-03-03
Add example number to comment list.

AaRaBiNoZa:
This is a feature proposed in #1883 . It's a simple change but in my case it's really useful and helps with collaboration.


Hironsan:
Thanks!

Hironsan:
close #1883

PR #2204, opened 2023-06-06
[Enhancement] Support clone project

### LINKED ISSUE #1818: Add ability to clone a project
Feature description
---------
Add a button on the Projects page that allows the user to clone an existing project and specify which sections to duplicate (Guidelines, Labels, Data, etc) into a new project.

Use Case
----------
Often, I want to do a test run of annotations to check the efficacy of my class labels and instructions before annotating a larger dataset. 
There are times when I need to perform multiple test runs after changing instructions, class labels, or both. This is a very manual workflow of creating multiple, near duplicate projects. Having the ability to duplicate or clone an existing project, and then make edits manually would be an improvement. Currently, the only option in Doccano is to create a new project from scratch and manually copy/paste the Guidelines and re-upload a labels file (or manually enter in the labels a second time).

Env: MacOS

Hironsan:
Enable to clone the following data:

- project
- dataset (examples)
- label types

Resolve #1818

### DIFF
--- a/backend/projects/models.py
+++ b/backend/projects/models.py
@@ -1,4 +1,5 @@
 import abc
+import uuid
 
 from django.conf import settings
 from django.contrib.auth.models import User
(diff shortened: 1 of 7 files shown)


Hironsan:
Caution: The files are not copied. Thus if you delete some files in an original or cloned project, you cannot access the files in another one.

PR #2214, opened 2023-06-12
Update for using cloudstorage

iwsh:
Fix a bug and add a new environment to enable to use S3 and S3-compatible storages in tasks involving image examples.


Hironsan:
Thanks!

PR #2296, opened 2023-10-26
fix: use unicode error symbol � instead of failing to import data

spodlesny:
Encoding detection is not a silver bullet, and some files can be broken inside, which ends up failing to import the whole batch of files (due to one broken file). For some tasks (sentiment detection), missing character is worth tradeoff. For other use-cases, at least the issue with encoding is presented to user in standardized way and on front-end - without looking into code, like I had to.

Replacement of null character (`\x00`) for � is required to successfully save data into PostgreSQL.


spodlesny:
Let me know if I should open and new Issue in bug tracker first. I had to fix it for my use-case, so I wanted to share my fix with community in case other have the same issue.

PR #2454, opened 2026-01-11
Enforce numeric ids in URL patterns 

Lesego-1:
### Summary
This PR updates URL patterns to enforce numeric IDs using <int:…> converters. 

### Changes
- Updated route `tasks/status/<task_id>` → `tasks/status/<int:task_id>`
- Ensures that invalid URLs (non-numeric IDs) return 404 automatically

### Motivation
- Enforcing integer IDs improves URL validation, reduces bugs, and makes the code more readable and maintainable

### Notes
- No backward compatibility issues are expected since all task IDs are numeric (AutoField)
- Tests for URL patterns should continue to pass

### DIFF
--- a/backend/api/urls.py
+++ b/backend/api/urls.py
@@ -3,5 +3,5 @@ from django.urls import path
 from .views import TaskStatus
 
 urlpatterns = [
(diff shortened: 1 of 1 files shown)


Lesego-1:
Thanks for the clarification. I’ve updated the route to use <uuid:task_id> and pushed the fix.

